PR #12, opened 2017-10-17
Add resolvers and model for Extension model

drshrey:


plugins/codeamp/models/models.go
```diff
@@ -140,8 +140,12 @@ type ExtensionSpec struct {
...
-	ExtensionId uuid.UUID `json:"extensionId" gorm:"type:uuid"`
-	State       string    `json:"state"`
+
+type Extension struct {
+	Model           `json:",inline"`
+	ProjectId       uuid.UUID     `json:"projectId" gorm:"type:uuid"`
+	ExtensionSpecId uuid.UUID     `json:"extensionSpecId" gorm:"type:uuid"`
+	State           plugins.State `json:"state"`
+	Artifacts       string        `json:"artifacts"`
+	Created         time.Time     `json:"created"`
```
sasso: no need to manually track Created

plugins/codeamp/schema/resolvers/extension.go
```diff
@@ -0,0 +1,51 @@
...
+func (r *ExtensionResolver) State() string {
+	return string(r.Extension.State)
+}
+
+func (r *ExtensionResolver) Created() graphql.Time {
+	return graphql.Time{Time: r.Extension.Created}
+}
+
+func (r *ExtensionResolver) Artifacts() string {
+	return r.Extension.Artifacts
```
sasso: We could return a key value list here, so it would be easier to render in the interface

PR #31, opened 2017-12-18
Refactor Extension.FormSpecValues to be a map[string]interface and models v. to Jsonb

drshrey:


plugins/codeamp/schema/resolvers/extension.go
```diff
@@ -144,28 +150,36 @@ func (r *Resolver) CreateExtension(ctx context.Context, args *struct{ Extension
...
-			log.InfoWithFields("FormSpecValuesIsValid failed", log.Fields{
-				"err": err,
+			log.InfoWithFields("can't marshal formSpecValuesMap", log.Fields{
+				"formSpecValuesMap": formSpecValuesMap,
 			})
-			return nil, err
+			return nil, errors.New("Can't marshal FormSpecValues")
 		}
 
+		// validate from formSpec
```
drshrey: commenting out for now as we figure out how validation will work in refactored ExtensionSpec definitions.

PR #33, opened 2018-01-24
Fix bugs in actions when passing data to Extension/ ReleaseExtension events. Add Env. branches and rm dex from Docker configs

drshrey:


docker-compose.yml
```diff
@@ -18,20 +18,17 @@ services:
       - codeamp
   circuit:
     image: codeamp:v001
-    command: sh -c "/usr/local/bin/dex serve /etc/dex/dex.dev.yaml & reflex -c reflex.conf"
+    command: sh -c "reflex -c reflex.conf"
```
amoskyler: We should leave dex in the docker-compose for local development

plugins/codeamp/actions/actions.go
```diff
@@ -255,6 +283,8 @@ func (x *Actions) ExtensionCreated(extension *models.Extension) {
 		}
 	}
 
+	spew.Dump("SECRETS DONE!", secrets)
```
amoskyler: remove spew logging

plugins/codeamp/actions/actions.go
```diff
@@ -546,6 +576,8 @@ func (x *Actions) WorkflowExtensionsCompleted(release *models.Release) {
...
```
amoskyler: remove spew

plugins/kubernetes/deployments/util.go
```diff
@@ -285,6 +296,8 @@ func (x *Deployments) doDeploy(e transistor.Event) error {
 		}
 	}
 
+	spew.Dump("parsed kubeconfig", config)
```
amoskyler: remove spew

plugins/kubernetes/deployments/util.go
```diff
@@ -334,6 +347,12 @@ func (x *Deployments) doDeploy(e transistor.Event) error {
 		secretMap[key] = val.(string)
 	}
 
+	for _, secret := range reData.Release.Secrets {
+		secretMap[secret.Key] = secret.Value
+	}
+
+	spew.Dump("SECRETS CREATED", secretMap)
```
amoskyler: spew

plugins/kubernetes/deployments/util.go
```diff
@@ -385,6 +404,8 @@ func (x *Deployments) doDeploy(e transistor.Event) error {
 		MountPath: "/etc/secrets",
 		ReadOnly:  true,
 	})
+
+	spew.Dump("FILE SECRETS", reData.Release.Secrets)
```
amoskyler: remove spew

docker-compose.yml
drshrey: yeah, re-added it to pull from the codeamp/dex-infra repo

plugins/kubernetes/deployments/util.go
sasso: @drshrey you should be super careful with spew. If this would be merged we would be leaking all the secrets to the logs...

Makefile
```diff
@@ -1,6 +1,6 @@
 up:
 	docker-compose up -d redis postgres
-	docker-compose run --rm circuit go run main.go --config ./configs/circuit.dev.yml migrate
```
sasso: can we change this back to `go run main.go`

docker-compose.yml
sasso: why pull from codeamp/dex-infra?

docker-compose.yml
```diff
@@ -18,20 +18,17 @@ services:
...
     build:
       context: .
       dockerfile: Dockerfile.dev
       args:
         pull: 1
     ports:
       - "${DOCKER_COMPOSE_CIRCUIT_PORT}:3011"
-      - "${DOCKER_COMPOSE_DEX_APP_PORT}:5555"
```
sasso: dex should not be removed

plugins/codeamp/actions/actions.go
```diff
@@ -50,24 +52,32 @@ func (x *Actions) GitSync(project *models.Project) {
...
-			Protocol:      project.GitProtocol,
-			Branch:        project.GitBranch,
-			RsaPrivateKey: project.RsaPrivateKey,
-			RsaPublicKey:  project.RsaPublicKey,
-		},
-		From: hash,
+	// get branches of entire environments
+	environments := []models.Environment{}
+	if x.db.Find(&environments).RecordNotFound() {
+		log.Info("no envs found")
```
sasso: Are you missing a return here?

plugins/codeamp/actions/actions.go
```diff
@@ -100,6 +110,31 @@ func (x *Actions) GitCommit(commit plugins.GitCommit) {
...
+	}
+
+	if x.db.Where("project_id = ? and name = ?", project.ID, branch.Name).First(&gitBranch).RecordNotFound() {
+		gitBranch = models.GitBranch{
+			ProjectId: project.ID,
+			Name:      branch.Name,
+		}
+		x.db.Save(&gitBranch)
+	} else {
+		log.InfoWithFields("feature already exists", log.Fields{
```
sasso: Why are we talking about a feature here?

docker-compose.yml
drshrey: addressed

plugins/codeamp/models/models.go
```diff
@@ -31,8 +31,9 @@ type UserPermission struct {
 }
 
 type Environment struct {
-	Model `json:",inline"`
-	Name  string `json:"name"`
+	Model     `json:",inline"`
+	Name      string `json:"name"`
+	GitBranch string `json:"gitBranch"`
```
sasso: Why do we have strings and table with branches, should this be a ID ref?

plugins/codeamp/schema/resolvers/environment.go
```diff
@@ -51,7 +59,13 @@ func (r *Resolver) UpdateEnvironment(ctx context.Context, args *struct{ Environm
 	if r.db.Where("id = ?", args.Environment.ID).Find(&existingEnv).RecordNotFound() {
 		return nil, fmt.Errorf("UpdateEnv: couldn't find environment: %s", *args.Environment.ID)
 	} else {
+		branch := "master"
```
sasso: Cleaner
```if args.Environment.GitBranch != nil {
    existingEnv.GitBranch = *args.Environment.GitBranch
} else {
    existingEnv.GitBranch = "master"
}```

Makefile
drshrey: oh yeah, addressed

plugins/codeamp/actions/actions.go
drshrey: I don't think so because there could be logic in the future that should run regardless of whether there are envs found. This will be deprecated though once we switch over to project-env specific branches. Should that be included in this pr?

plugins/codeamp/actions/actions.go
drshrey: ah this should be branch

plugins/codeamp/models/models.go
drshrey: removing this logic to replace with project-env branches in this pr

drshrey: and yes, there should be an ID ref although we should still keep the string name for display

plugins/codeamp/actions/actions.go
sasso: then you should use else because range environments will newer execute if this condition is met

plugins/codeamp/actions/actions.go
drshrey: addressed

plugins/codeamp/schema/resolvers/environment.go
sasso: I would also like to see how you will change the branch on project level. I think you should use  opposite approach and get branch first from project and if that is not defined use global one that is master.

plugins/kubernetes/deployments/util.go
drshrey: addressed

plugins/kubernetes/deployments/util.go
drshrey: addressed

plugins/kubernetes/deployments/util.go
drshrey: addressed

plugins/codeamp/actions/actions.go
drshrey: sure, that sounds good

plugins/codeamp/schema/resolvers/environment.go
drshrey: Do we still bind branches to envs? If not, the default for all envs would just be master, right? Until a project specifies something else in settings

docker-compose.yml
drshrey: i moved dex into its own service and am pulling from codeamp/dex-infra

PR #36, opened 2018-01-29
Move bootstrap dir to root; mv dex into circuit container for local dev; fix dockerfile

drshrey:


.gitignore
```diff
@@ -38,3 +38,5 @@ rm.sh
 credentials/*
 
 assets/*
+configs/dex.dev.yml
+configs/*.dev.yml
```
amoskyler: line 41 and 42 are duplicates.

It also looks like dex.dev.yml is already commited, so this rule doesn't do anything to remove the file - is there a reason we want to ignore changes to that file?

drshrey: yeah, i''ll remove this duplicate line. Reason is .dev.yml files should be optional and the user should by default use the given .yml files. Is there a good way to omit the *.dev.yml files from the repo?

Issue #52, opened 2018-01-30
Implement "Rollback" for releases


drshrey:

drshrey:
sub-items:
- persist all env info and artifacts into release object

PR #61, opened 2018-02-01
Insert newly created release into ReleaseCreated action in rollback

drshrey:


drshrey:
closes https://github.com/codeamp/circuit/issues/52

PR #66, opened 2018-02-05
Make artifacts additive when updating + add functionality to install once type extensions n times

drshrey:


drshrey:
closes https://github.com/codeamp/circuit/issues/67 and https://github.com/codeamp/circuit/issues/68

PR #81, opened 2018-02-20
Add better error message when commit is not found in DockerBuilder

drshrey:


plugins/dockerbuilder/dockerbuilder.go
```diff
@@ -167,16 +167,22 @@ func (x *DockerBuilder) build(repoPath string, event plugins.ReleaseExtension, d
...
-		io.Copy(os.Stderr, gitArchiveErr)
-	}()
+	dockerBuildErr := bytes.NewBuffer(nil)
 
+	io.Copy(dockerBuildErr, gitArchiveErr)
 	io.Copy(dockerBuildIn, gitArchiveOut)
 
 	err = gitArchive.Wait()
 	if err != nil {
+		if string(dockerBuildErr.Bytes()) != "" {
```
sasso: Do we need this if? we are already returning the error later on...

sasso: ping

drshrey: closing this pr for now and making an issue, going to come back to this later

PR #86, opened 2018-02-23
Add user permissions resolvers

drshrey:


plugins/codeamp/schema/resolvers/user.go
```diff
@@ -28,6 +35,56 @@ func (r *Resolver) CreateUser(args *struct{ User *UserInput }) *UserResolver {
...
+func (r *Resolver) UpdateUserPermissions(args *struct{ UserPermissions *UserPermissionsInput }) ([]string, error) {
+	invalidPermissionsErr := "Input parameters do not match distinct user permission scopes."
+	invalidUserIdErr := "User ID is not a valid UUID."
+
+	// Check if UserId input is valid
+	userId := uuid.FromStringOrNil(args.UserPermissions.UserId)
+	if userId == uuid.Nil {
+		return nil, fmt.Errorf(invalidUserIdErr)
+	}
+	if r.db.Where("id = ?").Find(&models.User{}).RecordNotFound() {
```
sasso: There is no userId defined in this search

plugins/codeamp/schema/resolvers/user.go
```diff
@@ -28,6 +35,56 @@ func (r *Resolver) CreateUser(args *struct{ User *UserInput }) *UserResolver {
...
+	}
+	// Check that all the input Permissions params
+	// are valid and within the distinct set of permission values
+	// Query the set of permission values
+	// by getting all 'distinct' values from user_permissions table
+	var distinctUserPermissionSet []models.UserPermission
+	if r.db.Select("DISTINCT(value)").Find(&distinctUserPermissionSet).RecordNotFound() {
+		log.Info(invalidPermissionsErr)
+		return nil, fmt.Errorf(invalidPermissionsErr)
+	}
```
sasso: there should be an empty space after if statements

plugins/codeamp/schema/resolvers/user.go
```diff
@@ -28,6 +35,56 @@ func (r *Resolver) CreateUser(args *struct{ User *UserInput }) *UserResolver {
...
+	if r.db.Where("id = ?").Find(&models.User{}).RecordNotFound() {
+		return nil, fmt.Errorf(invalidUserIdErr)
+	}
+	// Check that all the input Permissions params
+	// are valid and within the distinct set of permission values
+	// Query the set of permission values
+	// by getting all 'distinct' values from user_permissions table
+	var distinctUserPermissionSet []models.UserPermission
+	if r.db.Select("DISTINCT(value)").Find(&distinctUserPermissionSet).RecordNotFound() {
+		log.Info(invalidPermissionsErr)
```
sasso: Why do we throw invalidPermissionsErr when fetching a list of distinct values? I don't think we need to check for RecordNotFound here

plugins/codeamp/schema/resolvers/user.go
```diff
@@ -28,6 +35,56 @@ func (r *Resolver) CreateUser(args *struct{ User *UserInput }) *UserResolver {
...
+		return nil, fmt.Errorf(invalidPermissionsErr)
+	}
+	// Loop through and see if any of the input params
+	// don't match the given values
+	var found bool
+	for _, inputPermission := range args.UserPermissions.Permissions {
+		found = false
+		for _, distinctUserPermission := range distinctUserPermissionSet {
+			if inputPermission == distinctUserPermission.Value {
+				found = true
```
sasso: We need to check if the current user has this permission that he is granting

plugins/codeamp/schema/resolvers/user.go
```diff
@@ -28,6 +35,56 @@ func (r *Resolver) CreateUser(args *struct{ User *UserInput }) *UserResolver {
...
+			}
+		}
+		if !found {
+			return nil, fmt.Errorf(invalidPermissionsErr)
+		}
+	}
+	// First, we delete all the user_permissions rows related to the user_id input
+	// Then, we create user_permissions rows for each of the Permissions inputs
+	var results []string
+	r.db.Delete("user_id = ?", args.UserPermissions.UserId)
```
sasso: We need to rethink this. Let's say someone granted this user a permission and I don't have it, we will delete it and we cannot grant it back. We should do some check before we delete the permission and leave permissions we cant grant/delete

plugins/codeamp/schema/resolvers/user.go
```diff
@@ -28,6 +35,56 @@ func (r *Resolver) CreateUser(args *struct{ User *UserInput }) *UserResolver {
...
+	for _, inputPermission := range args.UserPermissions.Permissions {
+		userPermissionRow := models.UserPermission{
+			UserId: userId,
+			Value:  inputPermission,
+		}
+		r.db.Create(&userPermissionRow)
```
sasso: This is ok but only if we checked before that we can grant this permission

plugins/codeamp/schema/resolvers/users.go
```diff
@@ -23,3 +23,20 @@ func (r *Resolver) Users(ctx context.Context) ([]*UserResolver, error) {
 
 	return results, nil
 }
+
+func (r *Resolver) UserPermissionsOptions(ctx context.Context) ([]string, error) {
+	if _, err := utils.CheckAuth(ctx, []string{"admin"}); err != nil {
```
sasso: This could be available for users too and we should probably filter and return only the permissions user can grant or take away

plugins/codeamp/schema/resolvers/users.go
```diff
@@ -23,3 +23,20 @@ func (r *Resolver) Users(ctx context.Context) ([]*UserResolver, error) {
 
 	return results, nil
 }
+
+func (r *Resolver) UserPermissionsOptions(ctx context.Context) ([]string, error) {
```
sasso: Shouldn't this accept a userId and return a list of permissions available and which one is enabled? 
ex:
```
{
    "admin": false,
    "user": true,
}
```

plugins/codeamp/schema/resolvers/user.go
drshrey: good catch, thanks

plugins/codeamp/schema/resolvers/user.go
drshrey: fixing

sasso:
Can you please rebase this commit and not do the merge

Issue #95, opened 2018-03-08
Add migration code for env key


drshrey:
When updating key, we should have a migration that updates all existing envs in the DB with a key. The key's value will be the env name.

PR #96, opened 2018-03-08
Add migration and modify init-codeamp-db.sh to include testdb

drshrey:


drshrey:
addresses https://github.com/codeamp/circuit/issues/95

plugins/codeamp/migrations.go
```diff
@@ -336,6 +336,24 @@ func (x *CodeAmp) Migrate() {
...
+				db.Find(&environments)
+				for _, env := range environments {
+					if env.Key == "" {
+						env.Key = env.Name
+					}
+					db.Save(&env)
+				}
+				return nil
+			},
+			Rollback: func(tx *gorm.DB) error {
```
sasso: shouldn't this be key column?

sasso: or empty because you didn't create anything here

drshrey: yeah, it should be key. we need to drop the key column in rollback (since the migration is creating the key column)

PR #108, opened 2018-03-19
Add default env support

drshrey:


drshrey:
addresses https://github.com/codeamp/circuit/issues/109

PR #149, opened 2018-04-20
Remove admin for querying Users

drshrey:


drshrey:
We use Users to show users length in Dashboard component of panel which all codeamp users will see

PR #163, opened 2018-04-30
Search for soft-deleted project extensions for uninstalled project extension artifacts

drshrey:


plugins/codeamp/resolvers/release.go
```diff
@@ -91,7 +91,7 @@ func (r *ReleaseResolver) Artifacts(ctx context.Context) (JSON, error) {
 		var _artifacts []transistor.Artifact
 
 		projectExtension := ProjectExtension{}
-		if r.DB.Where("id = ?", releaseExtension.ProjectExtensionID).Find(&projectExtension).RecordNotFound() {
+		if r.DB.Unscoped().Where("id = ?", releaseExtension.ProjectExtensionID).Find(&projectExtension).RecordNotFound() {
```
amoskyler: Using `Unscoped` suggests we are querying all records even if they have been soft deleted. I understand that, but I don't fully understand why that fixes blank pages issues when selecting releases

PR #164, opened 2018-05-01
Mark stale releases as failed and add a timeout option for project settings

drshrey:


amoskyler:
@drshrey is this still something we are going to want to merge?

PR #182, opened 2018-05-10
Add route53 dns lock in (HOTFIX)

drshrey:


plugins/codeamp/resolvers/mutation.go
```diff
@@ -1468,6 +1504,16 @@ func (r *Resolver) BookmarkProject(ctx context.Context, args *struct{ ID graphql
 	}
 }
 
+func GetProjectExtensionWithRoute53Subdomain(subdomain string, db *gorm.DB) ProjectExtension {
+	var existingProjectExtension ProjectExtension
+
+	if db.Where("custom_config ->> 'subdomain' = ?", subdomain).First(&existingProjectExtension).RecordNotFound() {
```
amoskyler: We need to check both subdomain and hosted zone id to ensure uniqueness.

Additionally, we should ensure the subdomain and hosted zone ID are both lowercase to guarantee we catch case sensitive typos.

PR #189, opened 2018-05-15
Add logging to github status

drshrey:


plugins/githubstatus/githubstatus.go
```diff
@@ -229,14 +251,20 @@ func (x *GithubStatus) Process(e transistor.Event) error {
...
 					failedEvent := e.Payload.(plugins.ReleaseExtension)
 					failedEvent.State = plugins.GetState("failed")
 					failedEvent.Action = plugins.GetAction("status")
-					failedEvent.StateMessage = fmt.Sprintf("Timeout: try again and check if builds are taking too long fome reason.")
+					failedEvent.StateMessage = timeoutErrMsg
 					x.events <- e.NewEvent(failedEvent, nil)
 					return nil
 				}
-				log.Info("Looping through again and checking statuses")
+				log.InfoWithFields("Looping through again and checking statuses", log.Fields{
```
aballman: Seems like a good candidate for log.DebugWithFields. I believe the log level is presently set to info so this would disappear if you did that. I'm okay with setting default log level to debug to preserve this stuff, especially since you're obviously doing it to enhance troubleshooting.

Going to open a PR later so we can control the log level from an env or the circuit.yml file.

plugins/githubstatus/githubstatus.go
```diff
@@ -150,7 +152,9 @@ func (x *GithubStatus) Process(e transistor.Event) error {
 			for {
 				req, err := http.NewRequest("GET", fmt.Sprintf("https://api.github.com/repos/%s/commits/%s/status", event.Release.Project.Repository, event.Release.HeadFeature.Hash), nil)
 				if err != nil {
-					log.Info(err.Error())
+					log.InfoWithFields(err.Error(), log.Fields{
```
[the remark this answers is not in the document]
drshrey: Yup, makes sense!

plugins/githubstatus/githubstatus.go
drshrey: Great idea, I can mark it here and change once your log level change is in.

PR #190, opened 2018-05-15
Pass target urls for status objects in github status

drshrey:


plugins/githubstatus/githubstatus.go
```diff
@@ -189,7 +189,25 @@ func (x *GithubStatus) Process(e transistor.Event) error {
...
+						statuses := statusBodyInterface.(map[string]interface{})["statuses"].([]interface{})
+						if len(statuses) > 0 {
+							statusEvent := e.Payload.(plugins.ReleaseExtension)
+							statusEvent.State = plugins.GetState("waiting")
+							statusEvent.Action = plugins.GetAction("status")					
+							statusEvent.StateMessage = "Successfully got build events. Waiting for builds to succeed."
+							newStatusEvent := e.NewEvent(statusEvent, nil)							
+							for _, status := range statuses {
+								// send back artifacts containing the build urls for each status object
+								newStatusEvent.AddArtifact(fmt.Sprintf("%s", status.(map[string]string)["context"]), status.(map[string]string)["target_url"], false)
```
sasso: We should also add extension status like: circleci_status and circleci_url, so we know what failed.

drshrey: 👍

PR #195, opened 2018-05-17
Collect artifacts when timeout occurs for githubstatus

sasso:


plugins/githubstatus/githubstatus.go
```diff
@@ -214,87 +214,72 @@ func (x *GithubStatus) Process(e transistor.Event) error {
...
+					log.ErrorWithFields(err.Error(), log.Fields{
+						"hash": payload.Release.HeadFeature.Hash,
+					})
+
+					payload.State = plugins.GetState("failed")
+					payload.Action = plugins.GetAction("status")
+					payload.StateMessage = err.Error()
+					evt = e.NewEvent(payload, nil)
+					x.events <- e.NewEvent(payload, err)
+					return nil
```
drshrey: shouldn't this be break?

PR #197, opened 2018-05-17
Create workflow release extensions first and then deployments

drshrey:


amoskyler:
per discussed, release extension sorting should be done in api return

Issue #194, opened 2018-05-17
Update Dependencies


aballman:
A few deps are out of date

```
PROJECT                                  CONSTRAINT     VERSION                                 REVISION  LATEST      PKGS USED
github.com/Azure/go-ansiterm             branch master  branch master                           d6e3b33   d6e3b33     2
github.com/Microsoft/go-winio            v0.4.7         v0.4.7                                  7da180e   v0.4.7      1
github.com/Nvveen/Gotty                  branch master  branch master                           cd52737   cd52737     1
github.com/aws/aws-sdk-go                v1.13.33       v1.13.33                                63f3950   v1.13.33    27
github.com/azhao1981/go-engine.io        branch master  branch master                           4f280f1   4f280f1     6
github.com/azhao1981/go-socket.io        branch master  branch master                           fbcb463   fbcb463     1
github.com/azhao1981/go-socket.io-redis  branch master  branch master                           63556a5   63556a5     3
github.com/bitly/go-simplejson           v0.5.0         v0.5.0                                  aabad6e   v0.5.0      1
github.com/bww/go-json                   branch master  branch master                           2855ca2   2855ca2     1
github.com/codeamp/logger                branch master  branch master                           3af44e5   3af44e5     1
github.com/codeamp/transistor            branch master  branch refactor/events-have-event-info  34f9ff3   34f9ff3     1
github.com/containerd/continuity         branch master  branch master                           c6cef34   c6cef34     1
github.com/coreos/go-oidc                branch v2      branch v2                               1180514   1180514     1
github.com/davecgh/go-spew               v1.1.0         v1.1.0                                  346938d   v1.1.0      1
github.com/docker/docker                 branch master  branch master                           d6706dd   9aac187     25
github.com/docker/go-connections         v0.3.0         v0.3.0                                  3ede32e   v0.3.0      1
github.com/docker/go-units               v0.3.3         v0.3.3                                  47565b4   v0.3.3      1
github.com/extemporalgenome/slug         branch master  branch master                           0320c85   0320c85     1
github.com/fsnotify/fsnotify             v1.4.7         v1.4.7                                  c282820   v1.4.7      1
github.com/fsouza/go-dockerclient        v1.2.0         v1.2.0                                  ca33ff2   v1.2.0      1
github.com/garyburd/redigo               v1.6.0         v1.6.0                                  a69d193   v1.6.0      2
[...]

PR #201, opened 2018-05-18
Events data removed from payload

### LINKED ISSUE #198: Refactor plugins to use new transistor event interface
Transistor's event interface is changing soon. Update the plugins / message transceivers to handle new interface.

https://github.com/codeamp/transistor/issues/17

aballman:
This is a pretty heavy change.

Please review:
* Github Status plugin
* Mutation.go
* CodeAmp Plugin

Still yet to do:
* Adjust the ORM model for removed fields (state/statemessage)
* Adjust the seed data to not include removed fields
* Another pass of updating NewEvent calls to use transistor's event.Create() / event.Status() / etc.
* Update vendor folder for latest transistor edits  (could be done as part of https://github.com/codeamp/circuit/issues/194)

Fixes #198

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -107,7 +107,7 @@
   branch = "refactor/events-have-event-info"
   name = "github.com/codeamp/transistor"
   packages = ["."]
-  revision = "6aef95d399b5c6298fe5f0a8fe17611599d1daa4"
+  revision = "805ee56a440c34f281908674f7e43eea7b007c01"
 
 [[projects]]
   branch = "master"
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -107,7 +107,7 @@
   branch = "refactor/events-have-event-info"
   name = "github.com/codeamp/transistor"
   packages = ["."]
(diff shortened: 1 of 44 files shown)


aballman:
I still need to update the names of the events subscriptions in some places.

sasso:
Can you also resolve the conflicts before I start my review

aballman:
@sasso You got ahead of me, you speedy man, you. Working on it.

sasso:
and --rebase because I think we already merged LOG_LEVEL :)

aballman:
@sasso
https://github.com/codeamp/circuit/blob/91b96f9d774ec88716be78f8d204427efb0b843d/plugins/codeamp/codeamp.go#L183

It looks like the codeamp plugin receives messages, and then using reflection tries to hand it off to a handler on the payload.

Why doesn't that go through the normal message bus as opposed to having codeamp be a secondary handler for events?

sasso:
We were playing with different options of responding to events. The process became really big and hard to manage so we broke it down to methods. We could also go back to the standard way or have a switch case that calls methods...

PR #223, opened 2018-05-28
TLC and fixes for tests

### LINKED ISSUE #218: Reduce Duplicated Configuration of Tests
Just about every test loads it's own config file (which is fine). The problem is that in every test we are placing nearly identical code in the SetupTest function.

Something similar to the following:
```
func (suite *EnvironmentTestSuite) SetupTest() {
	viper.SetConfigType("yaml")
	viper.SetConfigFile("../../../configs/circuit.test.yml")
	err := viper.ReadInConfig()
	if err != nil {
		log.Fatal(err.Error())
	}
	db, err := gorm.Open("postgres", fmt.Sprintf("host=%s port=%s user=%s dbname=%s sslmode=%s password=%s",
		viper.GetString("plugins.codeamp.postgres.host"),
		viper.GetString("plugins.codeamp.postgres.port"),
		viper.GetString("plugins.codeamp.postgres.user"),
		viper.GetString("plugins.codeamp.postgres.dbname"),
		viper.GetString("plugins.codeamp.postgres.sslmode"),
		viper.GetString("plugins.codeamp.postgres.password"),
	))
	if err != nil {
		log.Fatal(err.Error())
	}
	db.AutoMigrate(
		&resolvers.Environment{},
	)
	// clean, just in case
	db.Delete(&resolvers.Environment{})
	suite.Resolver = &resolvers.Resolver{DB: db}
}
```

Should be moved out into a function that can handle these things in a central place.

aballman:
* Consolidated plugins tests setup into a `test.SetupPluginTest` function.
This handles log level and log format now and loading/registering plugins for tests.
* Updated plugin tests to use `transistor.GetState` as opposed to `plugins.GetState`
* Transistor Update: GetTestEvent returns an error now instead of issuing a panic after a timeout
Tests can now handle the timeout situation on their own.

Fixes #218

### DIFF
--- a/plugins/events.go
+++ b/plugins/events.go
@@ -20,6 +20,7 @@ func GetEventName(s string) transistor.EventName {
 		"gitsync",
 		"gitsync:commit",
 		"heartbeat",
+		"release",
 		"project:dockerbuilder",
 		"release:dockerbuilder",
 		"route53",
--- a/plugins/events.go
+++ b/plugins/events.go
@@ -11,16 +11,18 @@ import (
 
 func GetEventName(s string) transistor.EventName {
 	eventNames := []string{
-		"kubernetes:deployment",
-		"kubernetes:loadbalancer",
-		"githubstatus",
+		"project:kubernetes:deployment",
+		"release:kubernetes:deployment",
+		"project:kubernetes:loadbalancer",
+		"release:kubernetes:loadbalancer",
+		"project:githubstatus",
+		"release:githubstatus",
 		"gitsync",
 		"gitsync:commit",
 		"heartbeat",
-		"dockerbuilder",
+		"project:dockerbuilder",
+		"release:dockerbuilder",
 		"route53",
-		"release",
-		"project",
 		"websocket",
 	}
 
--- a/test/test.go
+++ b/test/test.go
@@ -1,4 +1,4 @@
-package tests
+package test
 
 import (
 	"bytes"
--- a/test/test.go
+++ b/test/test.go
@@ -11,7 +11,7 @@ import (
 	"github.com/spf13/viper"
 )
 
-func SetupPluginTest(viperConfig []byte, creatorsMap map[string]transistor.Creator) (*transistor.Transistor, error) {
+func SetupPluginTest(viperConfig []byte) (*transistor.Transistor, error) {
 	setupViperConfig(viperConfig)
 
 	pluginConfig := viper.GetStringMap("plugins")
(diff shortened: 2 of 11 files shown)


plugins/events.go
```diff
@@ -11,16 +11,18 @@ import (
...
+		"project:githubstatus",
+		"release:githubstatus",
 		"gitsync",
 		"gitsync:commit",
 		"heartbeat",
-		"dockerbuilder",
+		"project:dockerbuilder",
+		"release:dockerbuilder",
 		"route53",
-		"release",
```
sasso: Isn't release still an independent event called from codeamp?

test/test.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/spf13/viper"
+)
+
+func SetupPluginTest(pluginName string, viperConfig []byte, creator transistor.Creator) (*transistor.Transistor, error) {
+	setupViperConfig(viperConfig)
+	transistor.RegisterPlugin(pluginName, creator)
+
+	config := transistor.Config{
+		Plugins:        viper.GetStringMap("plugins"),
+		EnabledPlugins: []string{pluginName},
```
sasso: We should allow the test to run multiple plugins. This will be needed for integration testing.

test/test.go
```diff
@@ -12,12 +12,22 @@ import (
 )
 
 func SetupPluginTest(pluginName string, viperConfig []byte, creator transistor.Creator) (*transistor.Transistor, error) {
+	pluginMap := map[string]transistor.Creator{pluginName: creator}
+	return SetupMultiPluginTest(viperConfig, pluginMap)
+}
+
+func SetupMultiPluginTest(viperConfig []byte, creatorsMap map[string]transistor.Creator) (*transistor.Transistor, error) {
```
sasso: Just curious why not always use the config and why passing a pluginName is needed if you are also passing viperConfig?

test/test.go
aballman: https://github.com/codeamp/circuit/blob/4cbb18d21c112c79201c9d24cb1796a2f848594e/test/test.go#L19

This work for you?

test/test.go
aballman: Because the existing tests had no regard for what was in the viperConfig such as in heartbeat test.
https://github.com/codeamp/circuit/blob/2fbc3908fe5e76758cfca016bbd06f1892511604/plugins/heartbeat/heartbeat_test.go#L26

You make a fair point though, this can be improved to just read the plugins from the config file. Once that happens I can flatten it back into one function and not have to worry about passing plugin names anywhere else.

I like this, will fix.

PR #225, opened 2018-05-31
CircleCI Tests

aballman:
Getting things configured for tests to run in-container on CircleCI

Running into issues with CircleCI since it is docker-inside-docker.
https://circleci.com/docs/2.0/building-docker-images/#mounting-folders
https://circleci.com/docs/2.0/executor-types/#using-machine


aballman:
**Note**
I have removed `Dockerfile.dev` and `circuit.dev.yml` in favor of using `Dockerfile` and `circuit.yml`.
Maintaining both files is redundant. If there are dev edits on the config files, just don't check them in.
@sasso @amoskyler @drshrey

PR #230, opened 2018-06-05
Use default image pull policy instead of PullAlways

ziru:


plugins/kubernetes/deployment.go
```diff
@@ -220,9 +220,8 @@ func genPodTemplateSpec(e transistor.Event, podConfig SimplePodSpec, kind string
 				v1.ResourceMemory: resource.MustParse(podConfig.Service.Spec.MemoryRequest),
 			},
 		},
-		ImagePullPolicy: v1.PullAlways,
```
ziru: Remove the use of `v1.PullAlways` policy, to shorten the time used to pull images from docker registry, especially when the number of pods is high.

https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2107-L2113
The default policy (without specifying), `Defaults to Always if :latest tag is specified, or IfNotPresent otherwise`, should be good enough for typical uses.

amoskyler: The reason the policy is set to `PullAlways` is to mitigate issues when re-building images with modified parameters. We tag docker images with the commit hash of the build, so even if build args change on a specific commit (which will change the contents of the docker image) they are tagged identically.

This can lead to issues where an image already exists on the local machine but the repository image has been modified.

amoskyler: This means that we can run into sporadic issues where we are running different containers on different nodes. To deal with this, we just always re-pull the image from the docker registry.

ziru: I see.  How about tagging by commit hash (for tracking purpose only) but deploy via a uuid tag or the sha256 digest (for deployment purpose).
It seems anti-pattern to deploy using a tag, which can possibly refer to multiple images (at different build times, say, changing the build args).

amoskyler: I like this idea - I think it gives us the best of both worlds.

PR #245, opened 2018-06-12
Add pagination for Projects, Releases, Services, Features and Secrets

drshrey:
Implement pagination for queries to reduce network response time and provide a layer for rendering paginated interfaces on the frontend.

## Sample Query:
```
query {
  releases(params:{ limit: 5 }){
    nextCursor
    count
    page
    entries {
      id
    }
  }
}
```

## Sample Response:
```
{
  "data": {
    "releases": {
      "nextCursor": "6c331f48-a6aa-4e08-86bb-b937253df2ca",
      "count": 186,
      "page": 1,
      "entries": [
        {
          "id": "e70e20e7-4264-4bb3-aee1-9c8245e661a2"
        },
        {
          "id": "074aedc4-c510-4b8f-97ef-83457683d328"
        },
        {
          "id": "6d9f6c64-e950-4787-8927-79aed1a92497"
        }
      ]
    }
  }
}
```


plugins/codeamp/resolvers/paginator.go
```diff
@@ -0,0 +1,349 @@
...
+			if idx == len(entries) - 1 {
+				return 0, fmt.Errorf("Could not find given cursor.")
+			}			
+		}
+	}
+
+	return cursorRowIdx, nil
+}
+
+func EntryHelper(params PaginatorInput, entries interface{}) (interface{}, error) {
```
amoskyler: If I understand the use of this function correctly, we pass an unfiltered list with a full list of DB results, and then filter them down to meet the `PaginatorInput` parameter.

What happens when we have 100k releases to parse through? This way of doing pagination queries essentially means we dump an entire table into memory every time we make a request. It seems like it would make a lot more sense to perform the query itself using `PaginatorInput` as the constraints. 

Additionally using this method would likely allow us to break individual resource queries into their own resolvers, using `PaginatorInput` as an input.

plugins/codeamp/db/project.go
```diff
@@ -67,46 +67,44 @@ func (r *ProjectResolver) CurrentRelease() (*ReleaseResolver, error) {
...
+	Params *model.PaginatorInput
+}) *ReleaseListResolver {
 	var rows []model.Release
-	var results []*ReleaseResolver
 
-	r.DB.Where("project_id = ? and environment_id = ?", r.Project.Model.ID, r.Environment.Model.ID).Order("created_at desc").Find(&rows)
-	for _, release := range rows {
-		results = append(results, &ReleaseResolver{DB: r.DB, Release: release})
-	}
+	query := r.DB.Where("project_id = ? and environment_id = ?", r.Project.Model.ID, r.Environment.Model.ID).Order("created_at desc").Find(&rows)
```
amoskyler: You are executing `.Find` here, but it should be handled within the ReleaseListResolver.

plugins/codeamp/db/query.go
```diff
@@ -0,0 +1,342 @@
...
+		return nil, err
+	}
+
+	var rows []model.Extension
+	var results []*ExtensionResolver
+
+	if args.EnvironmentID != nil {
+		r.DB.Where("extensions.environment_id = ?", args.EnvironmentID).Order(`
+			CASE extensions.type
+				WHEN 'workflow' THEN 1
```
amoskyler: I can't help shaking the feeling that this is a kind of hacky way of handling this. I would almost prefer to perform two queries. One for workflows, and one for deployments - appending the results after the fact. Thoughts @drshrey @aballman?

plugins/codeamp/db/project.go
drshrey: ah correct, thank you for the find 🕵️

plugins/codeamp/db/user_test.go
```diff
@@ -0,0 +1,153 @@
+package db_resolver_test
```
amoskyler: Did you create this file as well?

plugins/codeamp/db/project_test.go
```diff
@@ -0,0 +1,205 @@
+package db_resolver_test
```
amoskyler: This appears to exist in `plugins/project_test.go`

plugins/codeamp/db/query.go
drshrey: Hmm well we'd have to do 3 queries since we'd have to find ones that aren't workflow or deployment. I'm not sure if that solution is better than the current state. The query string could be put into a var since there is redundancy there

PR #249, opened 2018-06-17
Slack notifier plugin

### LINKED ISSUE #202: Integrate with Slack
Please consider an enhancement to integrate with Slack. I expect to be notified when deploys succeed and fail.

aballman:
Fixes #202


plugins/events.go
```diff
@@ -25,6 +26,12 @@ func GetEventName(s string) transistor.EventName {
 		"release:dockerbuilder",
 		"route53",
 		"websocket",
+		"slack",
```
aballman: You can get rid of the create/update/delete/status events.
In this context, the ':' is really just used as a namespace.

The event is 'slack' and then when the action is 'create' the event will auto append them to make 'slack:create'.
You really only need a 'slack' event and a 'slack:notify' event.

plugins/codeamp/codeamp.go
```diff
@@ -1131,6 +1136,14 @@ func (x *CodeAmp) ReleaseFailed(release *resolvers.Release, stateMessage string)
 		x.DB.Save(&re)
 	}
 
+	if x.DB.Where("id = ?", release.ProjectID).First(&project).RecordNotFound() {
+		log.WarnWithFields("project not found", log.Fields{
+			"release": release,
+		})
+	}
+
+	x.SendNotifications("FAILED", release, &project)
```
drshrey: Should releaseState be a type State instead of a regular string? Also, we shouldn't need to pass release state as a separate param since we can grab its state within the release.

plugins/events.go
amoskyler: okay, great to know. I'll clean that up

plugins/events.go
```diff
@@ -148,35 +165,49 @@ type ServiceSpec struct {
...
 }
 
+// ProjectExtension event data struct
 type ProjectExtension struct {
 	ID          string  `json:"id"`
 	Project     Project `json:"project"`
 	Environment string  `json:"environment"`
 }
 
+// NotificationExtension event data struct
```
drshrey: Rather than create a new struct, could we add `Release` into the ProjectExtension struct? I'm not sure if there's anything fundamentally different about a notification to justify a new struct.

plugins/events.go
amoskyler: Actually there are a couple of places I use something like: ```if e.Name != plugins.GetEventName("slack:notify") {
		return nil
	}```

plugins/slack/slack.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// Subscribe to events
+func (x *Slack) Subscribe() []string {
+	return []string{
+		"slack:create",
+		"slack:update",
+		"slack:delete",
+		"slack:status",
+		"slack:notify",
```
drshrey: What's the need for a new action i.e. `slack:notify`?

plugins/slack/slack.go
```diff
@@ -0,0 +1,158 @@
...
+
+	payload := slack.Message{
+		Text:      fmt.Sprintf("Installed slack webhook to %s/%s", ePayload.Environment, ePayload.Project.Repository),
+		UserName:  "Codeamp",
+		Channel:   fmt.Sprintf("#%s", channel),
+		IconEmoji: fmt.Sprintf(":rocket:"),
+	}
+
+	webHookErr := sendSlackMessage(webhook, payload)
+
```
drshrey: remove the extra space here

plugins/events.go
drshrey: We shouldn't need to add all the possible action combos with the supported plugins. `slack` should be enough. We should also remove `release:dockerbuilder` ->`dockerbuilder` and `project:kubernetes:deployment`->`kubernetes`. @aballman, confirming this makes sense

drshrey: We could either modify `GetEventName` or have a func similar to what we have in transistor that does a regex match with supported events

aballman: I talked with Saso about the naming thing.
It makes more sense visually, but there are certain listeners waiting for all `release` events and if it's not `release:dockerbuilder` then there's no way of receiving all `release` events.

plugins/slack/slack.go
aballman: Which space? This file passes gofmt. What don't you like @drshrey ?

plugins/slack/slack.go
aballman: I'm with @drshrey  on this one, but the reverse. What's `slack` being used for? I only see `slack:notify`. If you aren't creating a `slack` event directly you don't need to register it separately.

plugins/codeamp/codeamp.go
amoskyler: It could be, however I want the ability to configure the messages from within codeamp. So string makes the most sense.

plugins/events.go
amoskyler: We could, however there is a good chance notifications will diverge in the future. I don't think it makes sense to attach a Release to ProjectExtensions, as they are two distinct things, and ProjectExtension will never have releases associated to it.

plugins/codeamp/codeamp.go
drshrey: Makes sense, and this is a generic function usable for any resource, so it doesn't make sense to derive state from something specific like Release.

PR #252, opened 2018-06-20
Resolvers Test Coverage

aballman:
We need more test coverage on resolvers.


plugins/codeamp/graphql/secret_test.go
```diff
@@ -0,0 +1,190 @@
...
+	ts.cleanupSecretIDs = append(ts.cleanupSecretIDs, secretResolver.ID())
+
+	_ = secretResolver.ID()
+	_ = secretResolver.Key()
+	_ = secretResolver.Value()
+	_ = secretResolver.Scope()
+	_ = secretResolver.Project()
+	_ = secretResolver.User()
+	_ = secretResolver.Type()
+	_, _ = secretResolver.Versions()
```
drshrey: I'm a bit confused as to what's happening here. Why is assignment necessary?

plugins/codeamp/auth/auth.go
```diff
@@ -7,65 +7,57 @@ import (
...
-}
-
-func SetAuthEnabled(enabled bool) {
-	log.Warn(fmt.Sprintf("AUTHENTICATION %s", enabledMap[enabled]))
-	authDisabled = !enabled
-}
-
-func checkAuthentication(ctx context.Context, scopes []string) (string, error) {
+func getJWTToken(ctx context.Context) (*model.Claims, error) {
+	if ctx == nil || ctx.Value("jwt") == nil {
```
amoskyler: Does this throw a runtime error if ctx is nil? As `ctx.Value` doesn't exist if `ctx == nil`.

plugins/codeamp/codeamp.go
```diff
@@ -83,7 +83,8 @@ func (x *CodeAmp) GraphQLListen() {
 	fs := http.FileServer(http.Dir(path.Join(path.Dir(filename), "graphql/static/")))
 	http.Handle("/", fs)
 
-	http.Handle("/query", graphql_resolver.CorsMiddleware(x.Resolver.AuthMiddleware(&relay.Handler{Schema: x.Schema})))
+	middleware := graphql_resolver.Middleware{x.Resolver}
+	http.Handle("/query", middleware.Cors(middleware.Auth(&relay.Handler{Schema: x.Schema})))
```
amoskyler: What's the reason for changing from `CorsMiddleWare` to `Cors`, and `AuthMiddleware` to `Auth`? Are these equivalent statements?

plugins/codeamp/graphql/mutation.go
```diff
@@ -1480,6 +1472,9 @@ func ExtractArtifacts(projectExtension model.ProjectExtension, extension model.E
 	if projectExtension.Artifacts.RawMessage != nil {
 		err = json.Unmarshal(projectExtension.Artifacts.RawMessage, &existingArtifacts)
 		if err != nil {
+			log.Error(err.Error())
+			log.Info(projectExtension.Artifacts.RawMessage)
```
amoskyler: Does this info dump artifact data, including protected values?

plugins/codeamp/codeamp.go
aballman: https://blog.golang.org/package-names
```
Avoid stutter. Since client code uses the package name as a prefix when referring to the package contents, the names for those contents need not repeat the package name. The HTTP server provided by the http package is called Server, not HTTPServer. Client code refers to this type as http.Server, so there is no ambiguity.
```

Moved this functionality into an appropriate namespace `middleware` and removed the suffix to avoid stutter.

plugins/codeamp/graphql/mutation.go
aballman: Error should be fine, will remove rawmessage dump, was used for debugging.

plugins/codeamp/graphql/secret_test.go
aballman: I'm using it as a placeholder in the test as a reminder to include asserts for these fields. Its resolved in the mutations PR. I had to cut it off somewhere.

plugins/codeamp/codeamp.go
amoskyler: Ahh I understand.

plugins/codeamp/graphql/environment_test.go
```diff
@@ -30,144 +32,154 @@ func (suite *EnvironmentTestSuite) SetupTest() {
...
 
 /* Test successful env. creation */
 func (suite *EnvironmentTestSuite) TestCreateEnvironment() {
-	envInput := model.EnvironmentInput{
-		Name:      "test",
-		Key:       "foo",
-		IsDefault: true,
-		Color:     "color",
+	// Environment
+	suite.helper.CreateEnvironment(suite.T())
```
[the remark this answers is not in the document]
aballman: This test is solely for the creation of the environment itself, not the interface. That test is handled in TestEnvironmentInterface. Inside the helper function `CreateEnvironment` it uses the `suite.T()` to do an `assert.FailNow` if there was a problem calling the `CreateEnvironment` mutation.

plugins/codeamp/auth/auth.go
aballman: That statement relies on short-circuiting. It's not explicit that go uses short circuiting evaluation, but it does. https://kuree.gitbooks.io/the-go-programming-language-report/content/22/text.html

It won't ever get to the second half of the if (after the `||`) because the first half is true already. It would only evaluate the second part of the if statement if the first half resulted in a `false` or if the expression was an `&&` instead of an `||`

aballman:
Merging this since @drshrey approved (didn't mean to dismiss) and I addressed the rogue log statement.

PR #253, opened 2018-06-20
allow multi-install notification plugins

amoskyler:


amoskyler:
related: https://github.com/codeamp/panel/pull/178

PR #255, opened 2018-06-22
Expose pod info to running container for metrics reporting purposes

ziru:
Four env variables exposed:
- MY_NODE_NAME
- MY_POD_NAME
- MY_POD_NAMESPACE
- MY_SERVICE_NAME

The first three are exposed via Kubernetes downward API:
https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/

The last one is directly exposed as name-value pair.

### DIFF
--- a/plugins/kubernetes/deployment.go
+++ b/plugins/kubernetes/deployment.go
@@ -364,18 +364,8 @@ func (x *Kubernetes) doDeploy(e transistor.Event) error {
 			myEnvVars = append(myEnvVars, newEnv)
 		}
 	}
-
-	// create env var for pod ip, constructed from pod metadata on deploy
-	podIPEnvVar := v1.EnvVar{
-		Name: "POD_IP",
-		ValueFrom: &v1.EnvVarSource{
-			FieldRef: &v1.ObjectFieldSelector{
-				FieldPath: "status.podIP",
-			},
-		},
-	}
-
-	myEnvVars = append(myEnvVars, podIPEnvVar)
+	// expose pod details to running container via env variables
+	myEnvVars = x.exposePodInfoViaEnvVariable(myEnvVars)
 
 	// as Files
 	var volumeMounts []v1.VolumeMount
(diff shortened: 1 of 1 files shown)


plugins/kubernetes/deployment.go
```diff
@@ -504,14 +506,20 @@ func (x *Kubernetes) doDeploy(e transistor.Event) error {
 			return err
 		}
 
+		// expose service name via env variable
+		podEnvVars := append(myEnvVars, v1.EnvVar{
```
drshrey: What's the reason for creating a new variable `podEnvVars` instead of appending to existing `myEnvVars`?

ziru: `myEnvVars` contains the same env variables shared for all pods.
`podEnvVars` contains all env vars from `myEnvVars`, plus the pod-specific one `MY_SERVICE_NAME`.

plugins/kubernetes/deployment.go
```diff
@@ -1017,3 +1031,34 @@ func (x *Kubernetes) doDeploy(e transistor.Event) error {
 
 	return nil
 }
+
+func (x *Kubernetes) exposePodInfoViaEnvVariable(myEnvVars []v1.EnvVar) []v1.EnvVar {
+	myEnvVars = append(myEnvVars, v1.EnvVar{
+		Name: "MY_NODE_NAME",
```
amoskyler: I would prefer to drop the `MY` and simple name these `NODE_NAME`, etc

plugins/kubernetes/deployment.go
```diff
@@ -364,6 +364,8 @@ func (x *Kubernetes) doDeploy(e transistor.Event) error {
 			myEnvVars = append(myEnvVars, newEnv)
 		}
 	}
+	// expose pod details to running container via env variables
+	myEnvVars = x.exposePodInfoViaEnvVariable(myEnvVars)
 
 	// create env var for pod ip, constructed from pod metadata on deploy
 	podIPEnvVar := v1.EnvVar{
```
amoskyler: @ziru we already append a FieldRef env var to our env vars here. We should either follow the same notation, or combine this with the function you implemented.

ziru: @amoskyler Done with merging into the new function.

plugins/kubernetes/deployment.go
ziru: Rename the first 3 metrics to use prefix `KUBE_`, and the last one to `CODEAMP_SERVICE_NAME`

PR #257, opened 2018-06-26
Plugin Refactor Fixes

aballman:
There were some issues that slipped under the radar.
This PR should address the issues discovered

### DIFF
--- a/plugins/kubernetes/deployment.go
+++ b/plugins/kubernetes/deployment.go
@@ -10,7 +10,6 @@ import (
 	"github.com/codeamp/circuit/plugins"
 	log "github.com/codeamp/logger"
 	"github.com/codeamp/transistor"
-	"github.com/davecgh/go-spew/spew"
 
 	apis_batch_v1 "k8s.io/api/batch/v1"
 	"k8s.io/api/core/v1"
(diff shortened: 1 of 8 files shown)


plugins/kubernetes/deployment.go
```diff
@@ -259,6 +260,8 @@ func genPodTemplateSpec(e transistor.Event, podConfig SimplePodSpec, kind string
 
 func (x *Kubernetes) doDeploy(e transistor.Event) error {
 	log.Info(e.Payload)
+	spew.Dump(e)
```
amoskyler: please remove

plugins/kubernetes/deployment.go
```diff
@@ -405,6 +408,8 @@ func (x *Kubernetes) doDeploy(e transistor.Event) error {
 	batchv1DepInterface := clientset.BatchV1()
 
 	// Validate we have some services to deploy
+	spew.Dump(reData)
```
amoskyler: remove these as well

plugins/slack/slack.go
```diff
@@ -18,7 +18,7 @@ type Slack struct {
 func init() {
 	transistor.RegisterPlugin("slack", func() transistor.Plugin {
 		return &Slack{}
-	}, plugins.NotificationExtension{})
+	}, plugins.NotificationExtension{}, plugins.ProjectExtension{})
```
amoskyler: Is this necessary? I think I may have missed the point of registering multiple extension types

aballman: It is for tests. Basically, you need to tell transistor how to marshal these structs. When running the application, the registration for `ProjectExtension` is handled elsewhere. Once they are isolated in the tests, the registration doesn't happen.

The tests rely on the plugin they are testing to do this registration.

PR #264, opened 2018-06-30
Mutation Tests

aballman:
A continuation of the resolvers test-coverage branch.


plugins/codeamp/graphql/mutation.go
```diff
@@ -902,7 +935,8 @@ func (r *Resolver) DeleteEnvironment(ctx context.Context, args *struct{ Environm
 			}
 
 			r.DB.Where("environment_id = ?", existingEnv.Model.ID).Delete(model.Release{})
-			r.DB.Where("environment_id = ?", existingEnv.Model.ID).Delete(model.ReleaseExtension{})
+			// ADB ReleaseExtension does not hae an environment ID
```
amoskyler: can we remove these?

plugins/codeamp/graphql/mutation.go
```diff
@@ -105,29 +104,25 @@ func (r *Resolver) CreateProject(ctx context.Context, args *struct {
...
+	environment := model.Environment{}
+	if r.DB.Where("id = ?", *args.Project.EnvironmentID).Find(&environment).RecordNotFound() {
+		log.InfoWithFields("Environment doesn't exist.", log.Fields{
 			"args": args,
 		})
-		return nil, fmt.Errorf("No envs found")
+		return nil, fmt.Errorf("Invalid Environment ID Provided")
 	}
 
-	for _, environment := range environments {
```
drshrey: We should be querying all environments and ignore the EnvironmentID param. See https://github.com/codeamp/circuit/pull/279

plugins/codeamp/graphql/mutation.go
aballman: We certainly can for ReleaseExtension. I think Release still needs it though, no?

amoskyler: oh sorry, I just meant the commented out code on 939 - i missed that the first was a real comment. It does have a typo in it though 🤔

PR #281, opened 2018-07-11
Restored old way of doing auth failures

aballman:


plugins/codeamp/graphql/middleware_test.go
```diff
@@ -61,21 +61,21 @@ func (ts *MiddlewareTestSuite) GetTestHandler() http.HandlerFunc {
 	return http.HandlerFunc(fn)
 }
 
-func (ts *MiddlewareTestSuite) TestAuthFailure() {
-	testserver := httptest.NewServer(ts.Middleware.Auth(ts.GetTestHandler()))
-	defer testserver.Close()
+// func (ts *MiddlewareTestSuite) TestAuthFailure() {
```
jperichon: should we just delete the code?

aballman: I want to save the test for when we have auth working correctly. I don't want to rewrite it.
Panel isn't handling 403 status well.

amoskyler: We could, but we will be re-introducing this code shortly, after we handle 40Xs properly.

PR #288, opened 2018-07-14
Service deploy strategy

### LINKED ISSUE #273: Allow to choose rollout strategy
**Is your feature request related to a problem? Please describe.**
The default rollout strategy is RollingUpdate, which is a decent choice.  For some services, I would want the Recreate strategy, so that all existing Pods are killed before new ones are created.

https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#writing-a-deployment-spec

Example services for Recreate:
- Kafka-based consumer services
RollingUpdate could cause more rebalancing than unnecessary.

- dd-agent services
RollingUpdate could cause 2x metrics at peak time.

- single pod service which assume no more than 1 pod instances
RollingUpdate could cause 2 pod instances during re-deploy.  The service logic then could need to lock to guarantee sequential processing, since there could be more than 1 pods.

amoskyler:
fixes #273


amoskyler:
yes

plugins/kubernetes/deployment.go
```diff
@@ -186,6 +186,53 @@ func detectPodFailure(pod v1.Pod) (string, bool) {
...
+				StrVal: "30%",
+			},
+			MaxSurge: &intstr.IntOrString{
+				Type:   intstr.String,
+				StrVal: "60%",
+			},
+		},
+	}
+
+	deploymentStrategy := plugins.DeploymentStrategy{}
```
drshrey: you don't need a var since you're using `plugins.DeploymentStrategy` once

plugins/codeamp/graphql/mutation.go
```diff
@@ -690,14 +694,23 @@ func (r *Resolver) CreateService(args *struct{ Service *model.ServiceInput }) (*
 		return &ServiceResolver{}, err
 	}
 
+	var deploymentStrategy model.ServiceDeploymentStrategy
+	if args.Service.DeploymentStrategy != nil {
+		deploymentStrategy, err = validateDeploymentStrategyInput(args.Service.DeploymentStrategy)
+		if err != nil {
+			return &ServiceResolver{}, err
```
aballman: I know this is littered about, but I would love it if we could get in the habit of doing 
`return nil, err` instead of an empty object. There may (or may not be) a perf impact from creating an empty object. Even if there isn't, I personally think it's more clear.

plugins/kubernetes/deployment.go
```diff
@@ -651,19 +698,7 @@ func (x *Kubernetes) doDeploy(e transistor.Event) error {
...
-						Type:   intstr.String,
-						StrVal: "30%",
-					},
-					MaxSurge: &intstr.IntOrString{
-						Type:   intstr.String,
-						StrVal: "60%",
-					},
-				},
-			}
+			deployStrategy = getDeploymentStrategy(service)
```
drshrey: these can go out of the if statements since it's the same line in either case

plugins/codeamp/db/service.go
```diff
@@ -67,6 +67,21 @@ func (r *ServiceResolver) Ports() ([]*model.JSON, error) {
 	return results, nil
 }
 
+// DeploymentStrategy
+func (r *ServiceResolver) DeploymentStrategy() (*model.JSON, error) {
```
drshrey: why do we not create the relevant graphql structures?

PR #292, opened 2018-07-14
Test Coverage: Pt 3

aballman:


plugins/codeamp/graphql/mutation.go
```diff
@@ -1288,40 +1289,40 @@ func (r *Resolver) UpdateProjectExtension(args *struct{ ProjectExtension *model.
...
+		return nil, fmt.Errorf("No extension found.")
 	}
 
 	project := model.Project{}
 	if r.DB.Where("id = ?", args.ProjectExtension.ProjectID).Find(&project).RecordNotFound() {
 		log.InfoWithFields("no project found", log.Fields{
 			"id": args.ProjectExtension.ProjectID,
 		})
-		return nil, errors.New("No project found.")
+		return nil, fmt.Errorf("No project found.")
```
drshrey: What's the reason for changing errors.New to fmt.Errorf?

plugins/codeamp/graphql/mutation.go
```diff
@@ -105,7 +108,7 @@ func (r *Resolver) CreateProject(ctx context.Context, args *struct {
 
 	// Create git branch for env per env
 	environments := []model.Environment{}
-	if r.DB.Find(&environments).RecordNotFound() {
+	if r.DB.Find(&environments).RecordNotFound() || len(environments) == 0 {
```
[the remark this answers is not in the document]
aballman: Technically, I should move the Find query out of the if statement and only check the length of environments. When you use the array/slice form of Find() it merely returns you an empty list if there are no records returned. You only receive an error if you use the struct version of Find and that's when you get a RecordNotFound error.

plugins/codeamp/graphql/mutation.go
```diff
@@ -1392,14 +1393,13 @@ func (r *Resolver) DeleteProjectExtension(args *struct{ ProjectExtension *model.
...
-		return &ProjectExtensionResolver{}, nil
-	}
-
-	for _, re := range res {
+	// ADB
+	// This was adjusted because ReleaseExtensions do not have an 'extension_id'
+	// Also, with gorm, when asking for a slice the Find() function will NOT
+	// return a record not found error. This only occurs when you are finding a specific
+	// record using Find(struct) as opposed to Find(slice)
+	r.DB.Where("project_extension_id = ?", args.ProjectExtension.ID).Find(&releaseExtensions)
```
drshrey: Should we check for `.Error` in this case to account for invalid args.ProjectExtension.ID errors?

plugins/codeamp/graphql/mutation.go
aballman: Consistency. We don't do strings.New("") we do fmt.Sprintf. It feels weird doing errors.New.

plugins/codeamp/graphql/extension_test.go
```diff
@@ -86,6 +86,16 @@ func (ts *ExtensionTestSuite) TestUpdateExtensionFailureNoEnv() {
...
 }
 
+func (ts *ExtensionTestSuite) TestUpdateExtensionFailureMissingEnv() {
+	// Update Extension
+	extensionInput := model.ExtensionInput{
+		ID: nil,
+	}
+
+	_, err := ts.Resolver.UpdateExtension(&struct{ Extension *model.ExtensionInput }{&extensionInput})
+	assert.NotNil(ts.T(), err)
```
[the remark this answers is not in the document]
aballman: Yes, the individual failure cases should have their own unique message.
This check is making sure that it was an error. There's no way to determine if the error is the one we want programmatically. I could do a string comparison against the error message but that doesn't work in the case of dynamic messages (something sprintf'd). It would also be very fragile as if anyone ever changed the text of the message it would break the string comparison.

plugins/codeamp/graphql/mutation.go
aballman: We check for validity of that at the beginning of the function 
```
	if r.DB.Where("id = ?", args.ProjectExtension.ID).First(&projectExtension).RecordNotFound() {
		log.InfoWithFields("no project extension found", log.Fields{
			"extension": args.ProjectExtension,
		})
		return nil, fmt.Errorf("No Project Extension Found")
	}
```

plugins/codeamp/graphql/mutation.go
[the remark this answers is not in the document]
aballman: The error will be caught by virtue of the following check that returns an error if environments is empty (whether by error or by legitimate lack of associated environments)

plugins/codeamp/graphql/extension_test.go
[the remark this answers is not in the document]
drshrey: In this case, it looks like we're only returning failure messages CodeAmp returns. If a dev working on CodeAmp changes the text of an error message, then wouldn't it be logical for them to reflect that change in the tests.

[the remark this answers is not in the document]
aballman: The test doesn't need a guarantee that the error is the error it is expecting.
If that was the case, then every error should be quantified and have a unique error code.

You can see here https://codecov.io/gh/codeamp/circuit/src/test-coverage-pt3/plugins/codeamp/graphql/mutation.go#L1143 that all cases are covered.

If someone is writing a PR and they notice test coverage goes down for a certain file, that should be investigated.

plugins/codeamp/graphql/mutation.go
drshrey: Got it, makes sense!

plugins/codeamp/graphql/mutation.go
drshrey: True, thoughts on removing that on adding `.Error` if it'll make it cleaner? This way, we can get more granular errors from GORM on why the parameter is wrong other than just saying "no project extension found". e.g. the id arg is an integer instead of a string

aballman: I am planning on addressing those in a separate PR.
This one is only meant to increase test coverage.

drshrey: Sounds good!

PR #296, opened 2018-07-18
Authentication

aballman:
* Overhauled auth process to handle hierarchial permissions without transistor. 
* Added auth tests. 
* Updated user query to only require auth if you're asking for a user's detail that is not the requester.

### DIFF
--- a/plugins/codeamp/auth/auth.go
+++ b/plugins/codeamp/auth/auth.go
@@ -27,14 +27,38 @@ func isAdmin(claims *model.Claims) bool {
 }
 
 func hasScopePermission(scope string, permissions []string) bool {
-	// Loop through each permission and see if it is a prefix
-	// of the scope we desire access to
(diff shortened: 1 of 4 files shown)


plugins/codeamp/auth/auth.go
```diff
@@ -26,25 +26,53 @@ func isAdmin(claims *model.Claims) bool {
...
+	// Permissions are built on a hierarchial basis with each tier being the string
+	// between two tokens '/' ex: "projects/codeamp/circuit" would cover permissions for that project
+	scopeTokens := strings.Split(scope, "/")
+	for _, permission := range permissions {
+		permissionTokens := strings.Split(permission, "/")
+
+		// If the permission hierarchy is deeper
+		// than the requested scope skip processing this one
+		// because the permission is more specific than the requested scope
+		if len(permissionTokens) > len(scopeTokens) {
```
drshrey: 👏

PR #301, opened 2018-07-20
Make credentials-related artifacts inaccessible to non-admins

drshrey:


plugins/dockerbuilder/dockerbuilder.go
```diff
@@ -466,10 +466,10 @@ func (x *DockerBuilder) Process(e transistor.Event) error {
...
 
 		ev := e.NewEvent(transistor.GetAction("status"), transistor.GetState("complete"), "Completed")
-		ev.AddArtifact("user", user.String(), user.Secret)
-		ev.AddArtifact("password", password.String(), password.Secret)
-		ev.AddArtifact("email", email.String(), email.Secret)
-		ev.AddArtifact("host", registryHost.String(), registryHost.Secret)
+		ev.AddArtifact("user", user.String(), true)
+		ev.AddArtifact("password", password.String(), true)
+		ev.AddArtifact("email", email.String(), true)
+		ev.AddArtifact("host", registryHost.String(), true)
```
amoskyler: Can you explain why we aren't proxying the artifacts own `secret` status down to this event's artifacts?

In a case like this, only password really needs to be secret (unless the admin created it as a secret env var)

jperichon:
Closing until a better solution

Issue #259, opened 2018-06-28
Write finished time for releases in DB


aballman:
There is a field called `finished_at` that is currently unused.
Start writing this value when releases are complete.

PR #313, opened 2018-07-31
Save finished_at times for release and release extension completions

drshrey:


jperichon:
closes https://github.com/codeamp/circuit/issues/259

PR #321, opened 2018-08-30
Fix rollbacks in kubernetes plugin

drshrey:


plugins/kubernetes/deployment.go
```diff
@@ -647,7 +647,9 @@ func (x *Kubernetes) doDeploy(e transistor.Event) error {
 	for _, service := range reData.Release.Services {
 		if service.Type == "one-shot" && !reData.Release.IsRollback {
 			oneShotServices = append(oneShotServices, service)
-		} else {
+		}
+
+		if service.Type == "general" {
```
amoskyler: can you add error handling in the case of a malformed service.Type?

PR #325, opened 2018-09-04
Make project query environment param optional

aballman:
* Makes the Environment ID parameter optional.

### DIFF
--- a/plugins/codeamp/graphql/project_query.go
+++ b/plugins/codeamp/graphql/project_query.go
@@ -50,32 +50,32 @@ func (u *ProjectResolverQuery) Project(ctx context.Context, args *struct {
 		return nil, err
 	}
 
-	if args.EnvironmentID == nil {
-		return nil, fmt.Errorf("Missing environment id")
-	}
+	if args.EnvironmentID != nil {	
 
-	environmentID, err := uuid.FromString(*args.EnvironmentID)
-	if err != nil {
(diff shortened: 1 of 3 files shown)


plugins/codeamp/graphql/project_query.go
```diff
@@ -50,32 +50,32 @@ func (u *ProjectResolverQuery) Project(ctx context.Context, args *struct {
...
-		log.InfoWithFields("Environment not found", log.Fields{
-			"environment": environmentID,
-			"identifier":  identifier,
-		})
-		return nil, fmt.Errorf("Environment not found")
-	}
+		// check if project has permissions to requested environment
+		var permission model.ProjectEnvironment
+		if u.DB.Where("project_id = ? AND environment_id = ?", resolver.DBProjectResolver.Project.Model.ID, environmentID).Find(&permission).RecordNotFound() {
+			log.InfoWithFields("Environment not found", log.Fields{
```
drshrey: Can we change this log statement to match the params? i.e.

```
log.InfoWithFields("ProjectEnvironment not found", log.Fields{
  "project_id": ...,
  "environment_id": ...
}
```

aballman: Can't provide project id because it may or may not be a project id that was used to identify the project. I updated the text to specify if it was the environment or project environment that failed to be found. Work?

PR #326, opened 2018-09-06
Fix CircleCI

aballman:
* Make CircleCI Fail Again (when tests fail)


aballman:
Going to just push this through as admin since it fixes the issues preventing tests from passing.

Issue #327, opened 2018-09-13
Disallow long service names


drshrey:
In relation to https://github.com/codeamp/panel/issues/251, we should also do server side validation in case the client side is either bypassed or a different frontend is used.

PR #329, opened 2018-09-13
Add service.Name validation so cannot be longer than 63 chars

drshrey:


drshrey:
fixes https://github.com/codeamp/circuit/issues/327

plugins/codeamp/graphql/mutation.go
```diff
@@ -712,6 +712,11 @@ func (r *Resolver) CreateRelease(ctx context.Context, args *struct{ Release *mod
 
 // CreateService Create service
 func (r *Resolver) CreateService(args *struct{ Service *model.ServiceInput }) (*ServiceResolver, error) {
+	// Check service name length
+	if len(args.Service.Name) > 63 {
+		return nil, errors.New("Service name cannot be longer than 63 characters.")
```
aballman: Can we switch to fmt.Errorf() instead of errors.New just for the sake of consistency?

plugins/codeamp/graphql/service_test.go
```diff
@@ -73,6 +73,36 @@ func (ts *ServiceTestSuite) TestCreateServiceSuccess() {
...
+
+	// Service Spec ID
+	serviceSpecResolver := ts.helper.CreateServiceSpec(ts.T())
+
+	// Deployment Strategy Input
+	deploymentStrategy := model.DeploymentStrategyInput{
+		Type: plugins.GetType("recreate"),
+	}
+
+	livenessProbe := model.ServiceHealthProbeInput{}
```
aballman: If you want to simplify this test, you can omit the liveness and readiness probes. They aren't required in CreateRelease. Same thing with deployment strategy.

PR #333, opened 2018-09-25
Disallows creation of multiple ingresses using the same FQDN for host…

amoskyler:
… routing

### DIFF
--- a/plugins/kubernetes/ingress.go
+++ b/plugins/kubernetes/ingress.go
@@ -217,6 +217,12 @@ func (x *Kubernetes) createIngress(e transistor.Event) error {
 
 	if inputs.Type == "loadbalancer" {
 
+		// check duplicates
+		isDuplicate, err := x.isDuplicateIngressHost(e)
+		if err != nil || isDuplicate {
+			return err
+		}
+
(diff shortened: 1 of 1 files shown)


plugins/kubernetes/ingress.go
```diff
@@ -294,6 +300,52 @@ func (x *Kubernetes) createIngress(e transistor.Event) error {
...
+		log.Error(err.Error())
+		return false, err
+	}
+
+	config, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: kubeconfig},
+		&clientcmd.ConfigOverrides{Timeout: "60"}).ClientConfig()
+
+	if err != nil {
+		failMessage := fmt.Sprintf("ERROR: %s; you must set the environment variable KUBECONFIG=/path/to/kubeconfig", err.Error())
```
drshrey: do we need to define `failMessage`? Can we omit this and just insert fmt.Sprintf in the log.Error?

amoskyler: good point, yes we can just omit it.

Issue #340, opened 2018-10-12
CreateRelease creates a release if it is not eligible


aballman:
**Describe the bug**
A release can be created if there are no valid release extensions. This leaves releases in an incompletable state.

PR #342, opened 2018-10-12
CreateRelease Creates Release if it is not Eligible

aballman:
Create Release will create a release if criteria are not matched.
Make sure release is only created after all the checks have been satisfied.


aballman:
Fixes #340

PR #345, opened 2018-10-15
patch: fix caching by getting consistent service signatures and add cacheable property 

drshrey:
Tested on dev and wrote unit tests for testing Cacheable property


drshrey:
Fixes https://github.com/codeamp/circuit/issues/346. We un-marshal services through a database query when we do a rollback; serviceID is omitted because it doesn't have a corresponding JSON tag. Because of that, we get a different signature which doesn't match with past releases' signatures. As a result, the cache condition does not execute.

drshrey:
@amoskyler it should not as service unit tests that are creating services with advanced settings are passing fine. i'm also testing in dev with deploy-test and advanced settings are being persisted

plugins/codeamp/migrations.go
```diff
@@ -120,10 +120,6 @@ func (x *CodeAmp) Migrate() {
 					"DOCKER_PASS",
 					"ACCESS_LOG_S3_BUCKET",
 					"SSL_CERT_ARN",
-					"CERTIFICATE_AUTHORITY",
```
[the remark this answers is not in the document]
drshrey: Yeah, we want to define them in `fileSecrets`, not in `envSecreets`. It was duplicated before, so yes, these were unused.

plugins/codeamp/release_handler.go
```diff
@@ -58,8 +58,13 @@ func (x *CodeAmp) ReleaseEventHandler(e transistor.Event) error {
 				eventState := transistor.GetState("waiting")
 				eventStateMessage := ""
 
-				// check if can cache workflows
-				if !release.ForceRebuild && !x.DB.Where("project_extension_id = ? and services_signature = ? and secrets_signature = ? and feature_hash = ? and state in (?)", projectExtension.Model.ID, releaseExtension.ServicesSignature, releaseExtension.SecretsSignature, releaseExtension.FeatureHash, []string{"complete"}).Order("created_at desc").First(&lastReleaseExtension).RecordNotFound() {
+				// query for the most recent complete release extension that has the same services, secrets and feature hash as this one
+				err = x.DB.Where("project_extension_id = ? and services_signature = ? and secrets_signature = ? and feature_hash = ? and state in (?)",
```
aballman: Can you bring this into the below if statement? If it's not cacheable forced rebuild there's no need to do the query.

plugins/codeamp/release_handler.go
```diff
@@ -58,8 +58,13 @@ func (x *CodeAmp) ReleaseEventHandler(e transistor.Event) error {
 				eventState := transistor.GetState("waiting")
 				eventStateMessage := ""
 
-				// check if can cache workflows
-				if !release.ForceRebuild && !x.DB.Where("project_extension_id = ? and services_signature = ? and secrets_signature = ? and feature_hash = ? and state in (?)", projectExtension.Model.ID, releaseExtension.ServicesSignature, releaseExtension.SecretsSignature, releaseExtension.FeatureHash, []string{"complete"}).Order("created_at desc").First(&lastReleaseExtension).RecordNotFound() {
+				// query for the most recent complete release extension that has the same services, secrets and feature hash as this one
```
aballman: Switched to requested change

plugins/codeamp/release_handler.go
drshrey: i can't because this query is part of what determines how we extract artifacts. for instance, even if an extension is cacheable and no force rebuilds, if no release extension is found, we still want to extract artifacts through `ExtractArtifacts`. we can achieve not executing the query but it would require duplicating that piece of code.

aballman: Why doesn't this work:
```				needsExtract := true
				if !release.ForceRebuild && extension.Cacheable {

					err = x.DB.Where("project_extension_id = ? and services_signature = ? and secrets_signature = ? and feature_hash = ? and state in (?)",
					projectExtension.Model.ID, releaseExtension.ServicesSignature,
					releaseExtension.SecretsSignature, releaseExtension.FeatureHash,
					[]string{"complete"}).Order("created_at desc").First(&lastReleaseExtension).Error

					if err != nil {
						eventAction = transistor.GetAction("status")
						eventState = lastReleaseExtension.State
						eventStateMessage = lastReleaseExtension.StateMessage

						err := json.Unmarshal(lastReleaseExtension.Artifacts.RawMessage, &artifacts)
						if err != nil {
							log.Error(err.Error())
							return nil
						}
						eventAction = transistor.GetAction("status")
						eventState = lastReleaseExtension.State
						eventStateMessage = lastReleaseExtension.StateMessage

						err = json.Unmarshal(lastReleaseExtension.Artifacts.RawMessage, &artifacts)
						if err != nil {
							log.Error(err.Error())
							return nil
						}

						needsExtract = false
					}
				}  

				if needsExtract == true {
					artifacts, err = graphql_resolver.ExtractArtifacts(projectExtension, extension, x.DB)
					if err != nil {
						log.Error(err.Error())
						return nil
					}
				}```

drshrey: OH yeah this works, 👍

PR #350, opened 2018-10-17
DockerBuilder Test Mocking

aballman:
Mocking up the DockerBuilder test with interfaces.

### DIFF
--- a/plugins/dockerbuilder/dockerbuilder.go
+++ b/plugins/dockerbuilder/dockerbuilder.go
@@ -23,11 +23,12 @@ import (
 type DockerBuilder struct {
 	events chan transistor.Event
 	Socket string
+	Dockerer
 }
 
 func init() {
 	transistor.RegisterPlugin("dockerbuilder", func() transistor.Plugin {
-		return &DockerBuilder{Socket: "unix:///var/run/docker.sock"}
+		return &DockerBuilder{Socket: "unix:///var/run/docker.sock", Dockerer: LegitimateDocker{}}
 	}, plugins.ReleaseExtension{}, plugins.ProjectExtension{})
 }
 
@@ -262,7 +263,7 @@ func (x *DockerBuilder) build(repoPath string, event transistor.Event, dockerBui
 		AuthConfigs:  authConfigs,
 	}
 
-	dockerClient, err := docker.NewClient(x.Socket)
+	dockerClient, err := x.Dockerer.NewClient(x.Socket)
 	if err != nil {
 		log.Error(err)
 		return err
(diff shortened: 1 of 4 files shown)


plugins/dockerbuilder/interfaces.go
```diff
@@ -0,0 +1,66 @@
...
+
+type Dockerer interface {
+	NewClient(Socket string) (DockerClienter, error)
+}
+
+type LegitimateDocker struct {
+}
+
+func (l LegitimateDocker) NewClient(socket string) (DockerClienter, error){
+	client, err := docker.NewClient(socket)
```
drshrey: if there's an err, we should return `nil, err`

plugins/dockerbuilder/interfaces.go
```diff
@@ -0,0 +1,66 @@
...
+	legitimateDockerClient := LegitimateDockerClient{ client }
```
drshrey: I know it's only one arg, but can we be explicit about what parameters we're setting in the struct?

plugins/dockerbuilder/dockerbuilder.go
```diff
@@ -380,7 +381,8 @@ func (x *DockerBuilder) getImageID(name string) (string, error) {
 		return "", err
 	}
 
-	truncatedID := image.ID[len(image.ID)-12:]
+	imageID := image.ID()
+	truncatedID := imageID[len(imageID)-12:]
```
drshrey: if imageID is only used once, it can just be `truncatedID := imageID[len(image.ID())-12:]`

plugins/dockerbuilder/interfaces.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func (l LegitimateDockerClient) PushImage(pushImageOptions docker.PushImageOptions, authConfig docker.AuthConfiguration) error {
+	return l.Client.PushImage(pushImageOptions, authConfig)
+}
+
+
+////////////////////////////
+
+type DockerImager interface {
```
drshrey: What's the need for an abstraction here? What's wrong with directly returning a `docker.Image` type?

aballman: Because when it's used it uses the variable ID directly. Interfaces can't define implementation details like variables, they just define which functions and what parameters they take. To get it to work in a test, I had to wrap the ID access with a function.

plugins/dockerbuilder/interfaces.go
aballman: docker.NewClient should already do that. This saves a branching statement.

plugins/dockerbuilder/dockerbuilder.go
aballman: In general I do this so that lines don't start to get complicated with a lot of operations happening all at once that you have to think about later when something broke. In this case, its imageID vs image.ID so I will drop it into one line for your nit.

aballman: Actually no, I'm keeping it. You missed the part where you referenced the variable you just cleaned up. In reality your version would be `truncatedID := image.ID()[len(image.ID())-12:]` and why call it twice. Also it's ugly.

aballman:
Admin merge due to Github webhooks issues.

PR #353, opened 2018-10-18
Kubernetes Tests

aballman:
* Restore old Kubernetes tests with mocked interfaces
* Refactored LoadBalancer and Deployments file a bit to break out larger function into smaller, more testable functions


aballman:
Tested on dev. Deployments and LoadBalancers deploy/add/delete successfully.

plugins/kubernetes/deployment_test.go
```diff
@@ -0,0 +1,198 @@
...
+				Spec: plugins.ServiceSpec{
+
+					CpuRequest:                    "10m",
+					CpuLimit:                      "500m",
+					MemoryRequest:                 "1Mi",
+					MemoryLimit:                   "500Mi",
+					TerminationGracePeriodSeconds: int64(1),
+				},
+				Replicas: 1,
+				Type:     "one-shot",
```
amoskyler: I haven't finished reading through the PR yet, but does it matter that we only test `one-shot` here?

plugins/kubernetes/ingress.go
```diff
@@ -196,7 +196,7 @@ func (x *Kubernetes) createIngress(e transistor.Event) error {
 		ingressRuleValue := v1beta1.IngressRuleValue{
 			HTTP: &v1beta1.HTTPIngressRuleValue{
 				Paths: []v1beta1.HTTPIngressPath{
-					v1beta1.HTTPIngressPath{
+					{
```
amoskyler: Neat, didn't know you could do this.

plugins/kubernetes/interfaces.go
```diff
@@ -0,0 +1,78 @@
...
+	contour_client "github.com/heptio/contour/apis/generated/clientset/versioned"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	v1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+///////////////////////////////////////////////////
```
amoskyler: What are the slashes for?

plugins/kubernetes/loadbalancer.go
```diff
@@ -8,12 +8,12 @@ import (
 	"time"
 
 	"k8s.io/api/core/v1"
-	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/codeamp/circuit/plugins"
 	log "github.com/codeamp/logger"
 	"github.com/codeamp/transistor"
+	_ "github.com/davecgh/go-spew/spew"
```
amoskyler: rm spew

plugins/kubernetes/testdata/kubeconfig
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+clusters:
+- cluster:
```
amoskyler: does this test actually use minikube to execute? You are using a k8s mock service, right?

plugins/kubernetes/deployment_test.go
aballman: I mean the long term goal is to have full code coverage, but this PR was meant just to get the tests running again with mocked interfaces.

plugins/kubernetes/ingress.go
aballman: I had a similar thought. `gofmt -s` performs the magic

plugins/kubernetes/interfaces.go
aballman: It's an empty comment block to help organize the structs. You don't care for?

plugins/kubernetes/testdata/kubeconfig
aballman: No, it doesn't. But if you don't provide the mock service with a valid kubeconfig it won't give you a mock connection. This is the bare minimum kubeconfig I could build that would be acceptable. We can rename it to something else if `minikube` is confusing.

I had to include it in testdata folder because on circleci or running somewhere that doesn't have one setup, the tests will fail.

plugins/kubernetes/interfaces.go
amoskyler: I don't find them particularly helpful, no.

plugins/kubernetes/testdata/kubeconfig
amoskyler: The localhost ip is what through me for a loop. Just wanted to confirm, as that IP is host/minikube dependent. I think it's fine as is.

PR #356, opened 2018-10-22
Slack Prettified

aballman:
* Broke report out into multiple lines
* Standardized icon used
* Included environment name in report


plugins/slack/slack.go
```diff
@@ -105,31 +105,46 @@ func (x *Slack) Process(e transistor.Event) error {
 	tail := payload.Release.TailFeature.Hash
 	head := payload.Release.HeadFeature.Hash
 
-	text := fmt.Sprintf(
-		"%s deployed <https://github.com/%s/compare/%s...%s|%s...%s> to <%s/projects/%s/%s/releases|%s>",
-		payload.Release.User, payload.Project.Repository, tail, head, tail[0:6], head[0:6], dashboardURL.String(), payload.Project.Slug, payload.Environment, payload.Project.Repository,
-	)
+	showGithubCompareUrl := true
```
drshrey: you don't need `showGithubCompareUrl`. It can just be 
```
// url compares two different commits to show diff
githubLinkURL := fmt.Sprintf("https://github.com/%s/commit/%s", payload.Project.Repository, head)
if tail == head {
  githubLinkURL = fmt.Sprintf("https://github.com/%s/compare/%s...%s", payload.Project.Repository, tail, head)
}
```

and just use githubLinkURL in slack.Attachment.Text.

plugins/slack/slack.go
```diff
@@ -105,31 +105,46 @@ func (x *Slack) Process(e transistor.Event) error {
...
+	}
+
+	// fmt.Sprintf("https://github.com/%s", payload.Project.Repository)
 
 	slackPayload := slack.Message{
-		Text:      text,
 		UserName:  "CodeAmp",
 		Channel:   fmt.Sprintf("#%s", channel.String()),
-		IconEmoji: resultEmoji,
+		IconEmoji: ":rocket:",
```
drshrey: we should probably have a standardized "codeamp" icon at some point but no need to this now

aballman: already brought it up with saso

Issue #359, opened 2018-11-01
Add Tests for DNS when Updating Extensions


aballman:
A user can select a domain in use by another project when creating a load balancer.
This will overwrite the existing DNS CNAME with a possibly incorrect load balancer endpoint.

PR #360, opened 2018-11-01
Fix route53 extensions same-subdomain prevention

drshrey:


aballman:
https://github.com/codeamp/circuit/issues/359

PR #361, opened 2018-11-05
Test to prevent duplicated Route53 entries

aballman:
Fixes https://github.com/codeamp/circuit/issues/359

### DIFF
--- a/plugins/codeamp/graphql/helpers_test.go
+++ b/plugins/codeamp/graphql/helpers_test.go
@@ -9,7 +9,7 @@ import (
 	"time"
 
 	db_resolver "github.com/codeamp/circuit/plugins/codeamp/db"
-	graphql_resolver "github.com/codeamp/circuit/plugins/codeamp/graphql"
+	. "github.com/codeamp/circuit/plugins/codeamp/graphql"
 	"github.com/codeamp/circuit/plugins/codeamp/model"
 	log "github.com/codeamp/logger"
 	"github.com/codeamp/transistor"
@@ -19,7 +19,7 @@ import (
 )
 
 type Helper struct {
-	Resolver *graphql_resolver.Resolver
+	Resolver *Resolver
 	name     string
 	context  context.Context
 
@@ -37,7 +37,7 @@ type Helper struct {
 	cleanupReleaseExtensionIDs          []uuid.UUID
 }
 
-func (helper *Helper) SetResolver(resolver *graphql_resolver.Resolver, name string) {
+func (helper *Helper) SetResolver(resolver *Resolver, name string) {
 	helper.Resolver = resolver
 	helper.name = name
 }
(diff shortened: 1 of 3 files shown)


plugins/codeamp/graphql/helpers_test.go
```diff
@@ -175,22 +199,25 @@ func (helper *Helper) CreateExtension(t *testing.T, envResolver *graphql_resolve
...
+func (helper *Helper) CreateExtensionWithInput(t *testing.T, extensionInput *model.ExtensionInput) *ExtensionResolver {
 	extensionResolver, err := helper.Resolver.CreateExtension(&struct {
 		Extension *model.ExtensionInput
-	}{Extension: &extensionInput})
+	}{Extension: extensionInput})
 	if err != nil {
 		assert.FailNow(t, err.Error())
 	}
 
-	assert.Equal(t, false, extensionResolver.Cacheable())
```
aballman: @drshrey I'm removing this from here. This should be part of testing the extension resolver interface, not in the helper function.

plugins/codeamp/graphql/helpers_test.go
```diff
@@ -247,7 +258,32 @@ func (helper *Helper) CreateProjectExtensionWithError(t *testing.T,
 	return projectExtensionResolver, err
 }
 
-func (helper *Helper) CreateFeature(t *testing.T, projectResolver *graphql_resolver.ProjectResolver) *graphql_resolver.FeatureResolver {
+func (helper *Helper) CreateProjectExtensionWithError(t *testing.T,
```
drshrey: Can you modify the commit message to reflect the changes other than just the test? Otherwise, these should be in a different PR.

plugins/codeamp/graphql/projectextension_test.go
```diff
@@ -495,6 +495,48 @@ func (ts *ProjectExtensionTestSuite) TestDeleteProjectExtensionFailureNoProjectE
...
+	extCustomConfigMap["listener_pairs"] = []string{}
+	extCustomConfigMap["subdomain"] = "test-subdomain"
+
+	_, err = ts.helper.CreateProjectExtensionWithConfig(ts.T(), extensionResolver, projectResolver, &extConfig, &extCustomConfigMap)
+	if err != nil {
+		assert.FailNow(ts.T(), err.Error())
+	}
+
+	_, err = ts.helper.CreateProjectExtensionWithConfig(ts.T(), extensionResolver, projectResolver, &extConfig, &extCustomConfigMap)
+	assert.NotNil(ts.T(), err)
```
drshrey: Since there are multiple failures that could arise from calling CreateProjectExtension, can we confirm the literal error string that is returned?

plugins/codeamp/graphql/helpers_test.go
aballman: Writing the test required new helper functions to be created. This is a prerequisite, and entirely related.

plugins/codeamp/graphql/projectextension_test.go
aballman: That would be a very fragile test with a string comparison. There are no other tests that check for the actual value of the error currently. If we want to do that, we would need to create an interface for errors that would include an error code that we could check against in addition to the string representation of the error.

I'm actually all for this, but it's outside the scope of this PR.

drshrey: I'm not sure how it would be fragile, but it makes sense to not include it in *this* PR. I've created an issue for it: https://github.com/codeamp/circuit/issues/365

PR #364, opened 2018-11-08
Make relationship between services and service specs 1-1

drshrey:


plugins/codeamp/migrations.go
```diff
@@ -443,7 +443,7 @@ func (x *CodeAmp) Migrate() {
 				return nil
 			},
 		},
-		// migrate ProjectExtension config to customConfig
+		// Add cacheable property to extensions
```
drshrey: not related but this was an incorrect comment that needed to be fixed

drshrey:
Partly addresses https://github.com/codeamp/circuit/issues/366

aballman:
This needs:
* A foreign-key constraint so that a service can't be created without an assigned service spec id. This also will prevent deleting service spec's accidentally without cleaning up the service
* A plan/script to remove the existing service specs so they aren't littered once we move to 1:1

Issue #366, opened 2018-11-08
Disallow service spec user selection


drshrey:
*Problem*: It's unclear how much resources should be allocated for services from a user-level. As a result, resources are either under or over-provisioned for services that cause the cluster they run on to be sub-optimal.

*Solution*: Remove the need for users to input that information and handle service-spec modifications on the admin side.

drshrey:
https://github.com/codeamp/panel/pull/279

PR #367, opened 2018-11-08
Add IsDefault property to ServiceSpecs

drshrey:


drshrey:
connected to https://github.com/codeamp/panel/compare/shreyas/add-default-servicespec?expand=1

drshrey:
partly addresses https://github.com/codeamp/circuit/issues/366

plugins/codeamp/graphql/servicespec_mutation.go
```diff
@@ -14,21 +14,34 @@ type ServiceSpecResolverMutation struct {
 	DB *gorm.DB
 }
 
+// CreateServiceSpec
 func (r *ServiceSpecResolverMutation) CreateServiceSpec(args *struct{ ServiceSpec *model.ServiceSpecInput }) (*ServiceSpecResolver, error) {
+
+	// if IsDefault is True, check which one is the current default
+	if args.ServiceSpec.IsDefault {
```
aballman: We should do these kinds of operations in a transaction when possible. Otherwise we risk the update being applied to one half and not the remaining half if there was an error

aballman:
What's the point of this? New services will need a service spec, but we've defined that we're going to have a 1:1 between services and specs moving forward. Does that mean the the designated 'default' is what the 'copies' are going to be created from? What happens if there is no default?

drshrey:
@aballman correct, we plan on always having a default profile which will be applied during the migration. Good point, there should never not be a default profile. I'll add logic to make sure there is always atleast one default in there, similar to how we do default environment.

aballman:
Have a look at http://gorm.io/docs/transactions.html

plugins/codeamp/graphql/servicespec_mutation.go
drshrey: good point, taking a look at the gorm docs now to see how to make this work

plugins/codeamp/db/servicespec.go
```diff
@@ -10,6 +13,32 @@ type ServiceSpecResolver struct {
 	DB *gorm.DB
 }
 
+// Name
+func (r *ServiceSpecResolver) Name() string {
+	env := model.Environment{}
+	project := model.Project{}
+	service := model.Service{}
+ 	if err := r.DB.Where("id = ?", r.ServiceSpec.ServiceID).First(&service).Error; err != nil {
+		log.InfoWithFields("service not found", log.Fields{
```
amoskyler: we should be a bit more thoughtful in how we handle these Errors. The implication is there is some sort of error, but isn't necessarily a NotFound error. We should both check for not found errors as well as logging the error if it's something unexpected.

drshrey: true, i'll replace the `service not found` string with the error itself. will do the same for the others in the function as well

aballman: I think @amoskyler meant more along these lines
```
if err := db.Where("name = ?", "jinzhu").First(&user).Error; gorm.IsRecordNotFoundError(err) {
  // record not found
}
```

Specifically the `gorm.IsRecordNotFoundError(err)` part.
http://gorm.io/docs/error_handling.html

aballman:
This decreases code coverage?

PR #368, opened 2018-11-13
Disable servicespec selection on service creation

drshrey:
This PR does a few things:
- Removes ServiceSpecID param from Service creation
- Generates a service spec that derives properties from default service spec, for each service
- makes service spec mutations atomic


plugins/codeamp/db/service.go
```diff
@@ -40,7 +40,7 @@ func (r *ServiceResolver) Name() string {
 func (r *ServiceResolver) ServiceSpec() *ServiceSpecResolver {
 	var serviceSpec model.ServiceSpec
 
-	r.DB.Model(r.Service).Related(&serviceSpec)
+	r.DB.Where("service_id = ?", r.Service.Model.ID).First(&serviceSpec)
```
aballman: Can we add some error handling here? At the very least we should log the error if there is one.

plugins/codeamp/graphql/project_mutation.go
```diff
@@ -116,6 +114,8 @@ func (r *ProjectResolverMutation) CreateProject(ctx context.Context, args *struc
 		return nil, fmt.Errorf("No envs found")
 	}
 
+	r.DB.Create(&project)
```
aballman: I don't mind if we do this in another PR, but we should probably add some error checking here as well on the create operation itself.

plugins/codeamp/graphql/project_test.go
```diff
@@ -276,6 +279,7 @@ func (suite *ProjectTestSuite) TestCreateProjectFailure() {
 	}
 
 	// Project
+	suite.helper.CreateProjectWithInput(suite.T(), &projectInput)
```
aballman: Is this intentional to create two here?

plugins/codeamp/graphql/service_mutation.go
```diff
@@ -90,7 +90,20 @@ func (r *ServiceResolverMutation) CreateService(args *struct{ Service *model.Ser
 		PreStopHook:        preStopHook,
 	}
 
-	r.DB.Create(&service)
+	r.DB.Create(&service)
```
aballman: Check for error on create

plugins/codeamp/graphql/project_mutation.go
drshrey: yup, created an issue here: https://github.com/codeamp/circuit/issues/380

PR #372, opened 2018-11-19
Kubernetes Tests

aballman:
* Restore old Kubernetes tests with mocked interfaces
* Refactored LoadBalancer and Deployments file a bit to break out larger function into smaller, more testable functions
* Added test to ensure secrets are provided to deploys

This branch was originally merged in with the PR #353.
There was a bug that broke secret creation for pods.
This has been resolved with commits added to this branch - and this is the replacement PR.


plugins/kubernetes/deployment_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+type TestSuiteDeployment struct {
+	suite.Suite
+	transistor *transistor.Transistor
+	MockBatchV1Job
+}
+
+func (suite *TestSuiteDeployment) SetupSuite() {
+	var viperConfig = []byte(`
+plugins:
```
drshrey: Not necessary in PR but we may want to look into creating yaml files with a go library and converting them to bytes rather than directly using literal strings

aballman: We could add the file in the `./testdata` directory and just have a config file per test, and load it into bytes like any other file.

Issue #406, opened 2019-01-25
Caching still broken


drshrey:
Caching is still broken for deploys. I expect docker images to be cached when I check the "caching" box in the extensions page for rollbacks and redeploys. Currently, images still rebuild on rollbacks

PR #408, opened 2019-01-26
Add caching fix with test case

drshrey:
Fixes https://github.com/codeamp/circuit/issues/406


plugins/codeamp/release_handler.go
```diff
@@ -66,9 +66,11 @@ func (x *CodeAmp) ReleaseEventHandler(e transistor.Event) error {
 						releaseExtension.SecretsSignature, releaseExtension.FeatureHash,
 						[]string{"complete"}).Order("created_at desc").First(&lastReleaseExtension).Error
 					if err != nil {
+						log.Error(err.Error())
```
amoskyler: You should return the error as well.

drshrey: not here because we still want to continue execution if a release extension isn't found

Issue #401, opened 2019-01-10
Export/Import of Environment Variables


aballman:
There is currently a very large problem when it comes to managing environment variables.
It's a slow, manual process that has to be completed for every environment and project that is created.

I propose that we relieve some of that pain by allowing an import/export of environment variables using  a .env style or some other key,value mapping. This would allow users to quickly and consistently work with environment variables across project scopes.

I'm imagining that in the secrets page, we could build in a button to accomplish this without rocking the boat too much in terms of env management.

PR #411, opened 2019-01-29
Add secrets importer

drshrey:
Fixes the import portion of this - https://github.com/codeamp/circuit/issues/401


drshrey:
@aballman thanks for the comment, totally missed out on validating `protected-env`s. It's addressed now though!

plugins/events.go
```diff
@@ -85,7 +85,7 @@ func GetType(s string) Type {
 		errMsg += fmt.Sprintf("%s : ln %d", file, line)
 	}
 
-	log.Panic(errMsg)
+	log.Warn(errMsg)
```
aballman: Would prefer to have this be an Error level message if not a panic.

plugins/codeamp/graphql/secret_test.go
```diff
@@ -211,6 +212,341 @@ func (ts *SecretTestSuite) TestSecretScopes() {
...
+	if err != nil {
+		assert.FailNow(ts.T(), err.Error())
+	}
+
+	assert.Equal(ts.T(), 2, len(secretsResolver))
+
+	// check that protected was created
+	protectedSecrets := []model.Secret{}
+	err = ts.Resolver.DB.Where("project_id = ? and environment_id = ? and is_secret = ?",
+		projectResolver.DBProjectResolver.Project.Model.ID.String(), envResolver.DBEnvironmentResolver.Environment.Model.ID.String(), true).Find(&protectedSecrets).Error
```
aballman: This is an acceptable means of testing for secrets, however it would be better if we used the Secrets resolver to accomplish this to ensure that the path the client is going to take accurately reflects the conditions of the test.

PR #413, opened 2019-01-29
Add secrets exporter

drshrey:
Fixes 50% of https://github.com/codeamp/circuit/issues/416


plugins/codeamp/graphql/secret_test.go
```diff
@@ -490,7 +490,7 @@ func (ts *SecretTestSuite) TestSecretsImport_Fail_InvalidSecretsType() {
 }
 
 func (ts *SecretTestSuite) TestSecretsImport_Success_ProtectedSecretCreated() {
-	ts.T().Log("TestSecretsImport_Fail_InvalidSecretsType")
+	ts.T().Log("TestSecretsImport_Success_ProtectedSecretCreated")
```
aballman: This isn't really needed since the test suite announces what tests its working on

plugins/codeamp/graphql/secret_query.go
```diff
@@ -42,3 +43,45 @@ func (r *SecretResolverQuery) Secret(ctx context.Context, args *struct {
...
 	return &resolver, nil
 }
+
+// ExportSecrets returns a list of all secrets for a given project and environment in a YAML string format
+func (r *SecretResolverQuery) ExportSecrets(ctx context.Context, args *struct{ Params *model.ExportSecretsInput }) (string, error) {
+	project := model.Project{}
+	env := model.Environment{}
+	secrets := []model.Secret{}
+
+	if err := r.DB.Where("id = ?", args.Params.ProjectID).First(&project).Error; err != nil {
```
aballman: I think it would be worthwhile to turn on db logging at some point and take a look at this query that's generated.
It's not an issue now, or for small scale things, but if `First` is asking the DB to `ORDER` and then `LIMIT` a simple `SELECT` should be fine when using a uid to select.

aballman: I won't reject over this

plugins/codeamp/graphql/secret_test.go
aballman: Remove and I'll approve

plugins/codeamp/graphql/secret_query.go
drshrey: It does:
```
SELECT * FROM "projects"  WHERE "projects"."deleted_at" IS NULL AND ((id = 'foo-id')) ORDER BY "projects"."id" ASC LIMIT 1
```

Yeah, not an issue now, but I changed it here to use `Find` instead. Later, I can make a PR to overhaul to Find where it's possible. Thanks for that observation. (issue for later - https://github.com/codeamp/circuit/issues/420)

PR #417, opened 2019-01-30
Add services exporter

drshrey:
Fixes https://github.com/codeamp/circuit/issues/401

### DIFF
--- a/plugins/codeamp/graphql/service_query.go
+++ b/plugins/codeamp/graphql/service_query.go
@@ -3,10 +3,12 @@ package graphql_resolver
 import (
 	"context"
 
+	"github.com/codeamp/circuit/plugins"
 	"github.com/codeamp/circuit/plugins/codeamp/auth"
 	db_resolver "github.com/codeamp/circuit/plugins/codeamp/db"
 	"github.com/codeamp/circuit/plugins/codeamp/model"
 	"github.com/jinzhu/gorm"
+	yaml "gopkg.in/yaml.v2"
 )
 
 // Service Resolver Query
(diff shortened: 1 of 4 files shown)


plugins/codeamp/graphql/service_query.go
```diff
@@ -30,3 +33,76 @@ func (r *ServiceResolverQuery) Services(ctx context.Context, args *struct {
...
+		if err := r.DB.Where("service_id = ?", service.Model.ID).First(&deployStrategy).Error; err != nil {
+			log.Info(err.Error())
+		}
+		service.DeploymentStrategy = deployStrategy
+
+		// Get readiness probe
+		readinessProbe := model.ServiceHealthProbe{}
+		rpHeaders := []model.ServiceHealthProbeHttpHeader{}
+		if err := r.DB.Where("service_id = ? and type = ?", service.Model.ID, string(plugins.GetType("readinessProbe"))).First(&readinessProbe).Error; err != nil {
+			log.Info(err.Error())
```
aballman: log as errors. if it's not important that the query find a valid result, then wrap it in a `if gorm.IsRecordNotFound(err) == false` and log the error there

plugins/codeamp/graphql/service_query.go
```diff
@@ -30,3 +33,76 @@ func (r *ServiceResolverQuery) Services(ctx context.Context, args *struct {
...
+		}
+
+		readinessProbe.HttpHeaders = rpHeaders
+		service.ReadinessProbe = readinessProbe
+
+		// get liveness probe
+		livenessProbe := model.ServiceHealthProbe{}
+		lpHeaders := []model.ServiceHealthProbeHttpHeader{}
+		if err := r.DB.Where("service_id = ? and type = ?", service.Model.ID, string(plugins.GetType("livenessProbe"))).First(&livenessProbe).Error; err != nil {
+			log.Info(err.Error())
```
aballman: log.error

plugins/codeamp/graphql/service_query.go
```diff
@@ -30,3 +33,76 @@ func (r *ServiceResolverQuery) Services(ctx context.Context, args *struct {
...
+		}
+
+		if err := r.DB.Where("health_probe_id = ?", livenessProbe.Model.ID).Find(&lpHeaders).Error; err != nil {
+			log.Info(err.Error())
```
aballman: log.error

Issue #369, opened 2018-11-14
 Have version consistency among services at all times


drshrey:
Currently, if n services go out during a deployment and x succeed but n -x fail, the service versions become split between the current deployment and previous deployment.

We should rollback to the previous version for any services that did succeed if the overall deployment step itself fails.

PR #418, opened 2019-01-31
Unwind Failed Deploys

aballman:
* Coordinates the release process between Kubernetes deployments. If one deploy fails, rollback all deployments to their previous version.
* Handles deployment failures on first time deployment by deleting all deploys/replicasets/pods
* DANGER: This does not handle rolling back migrations
Fixes https://github.com/codeamp/circuit/issues/369


plugins/kubernetes/deployment.go
```diff
@@ -1292,6 +1356,182 @@ func (x *Kubernetes) doDeploy(e transistor.Event) error {
...
+			if targetReplicaSet == nil {
+				log.Error(fmt.Sprintf("UNWIND-DEPLOY: Target Replica Set Not Found for: %s", deploymentName))
+				continue
+			}
+
+			// Update the deployment, change the spec to match
+			// the replicaset spec
+			deployment.Spec.Template = targetReplicaSet.Spec.Template
+			_, err = depInterface.Deployments(namespace).Update(deployment)
+			if err != nil {
```
drshrey: Why don't we need to check for a successful deploy here as well? When we update the spec, isn't it the same process of deploying a set of new pods?

drshrey:
@aballman do we need test cases to handle the new deployment logic? I ask because the behavior here is different as the end state is different (rollback on failure) from the current state of doDeploy

PR #423, opened 2019-02-05
Fix schema.graphql errors for Import/Export related functions

drshrey:
The `environmentID` was absent for `importSecrets` and the `userID` can be derived from context as opposed to the user having to pass it in (which makes the `userID` attribute unnecessary). This PR adds in the `environmentID` and removes the `userID`. The unit tests do not cover the validity of the GraphQL schema, so this was missed in those tests. In the future, I'll have to compensate for that with a more manual testing process or incorporating schema validity into our automated tests.


plugins/codeamp/graphql/secret_test.go
```diff
@@ -244,8 +244,12 @@ func (ts *SecretTestSuite) TestSecretsImport_Success() {
 	}
 
 	// call importer function
-	ctx := context.Background()
-	secretsResolver, err := ts.Resolver.ImportSecrets(ctx, &struct{ Secrets *model.ImportSecretsInput }{
+	userContext := context.WithValue(context.Background(), "jwt", model.Claims{
```
aballman: You could use https://github.com/codeamp/circuit/blob/master/test/test.go#L72 here

PR #427, opened 2019-02-07
Unwind Failed Deploys Improvements

aballman:
* Unwinds rollbacks with scaling operations
* Handles removing deploys/rs when they are a 'new' service part of an existing deploy
* Grabs ReplicaSet definitions before Deployment and stores in-memory
* Simplifies testing framework for Deployments test


plugins/kubernetes/deployment.go
```diff
@@ -1384,98 +1394,163 @@ func (x *Kubernetes) unwindFailedDeployment(clientset kubernetes.Interface, name
...
-				log.Error(fmt.Sprintf("UNWIND-DEPLOY: %s", err.Error()))
-				return err
 			}
 
-			// Delete the deployment
-			err = x.ExtDeploymenter.Delete(clientset, namespace, deploymentName, &meta_v1.DeleteOptions{TypeMeta: meta_v1.TypeMeta{
-				Kind: "Deployment",
-			}})
+			log.Warn(fmt.Sprintf("SCALING DOWN REPLICASET FROM %d REPLICAS", rs.Status.Replicas))
+			_, err = clientset.Extensions().ReplicaSets(namespace).UpdateScale(rs.Name, scale)
```
drshrey: What's the reason for updating the scale of orphaned replica sets as opposed to just outright deleting them? If they're orphaned, then why should we let them continue living in the cluster?

plugins/kubernetes/deployment_test.go
```diff
@@ -443,4 +468,69 @@ func BasicReleaseExtensionJobAndService() plugins.ReleaseExtension {
...
+			{
+				Key:   "secret-key",
+				Value: "secret-value",
+				Type:  plugins.GetType("internal"),
+			},
+		},
+	}
+
+	releaseExtension := plugins.ReleaseExtension{
+		//		Slug:    "kubernetesdeployments",
```
drshrey: nit: remove if not being used or justify the comment

plugins/kubernetes/deployment_test.go
```diff
@@ -228,11 +257,8 @@ func (suite *TestSuiteDeployment) TearDownSuite() {
 	suite.transistor.Stop()
 }
 
-func FailingReleaseExtension() plugins.ReleaseExtension{
-	extension := BasicReleaseExtensionJobAndService()
-	extension.Release.Services[0].Command = "/bin/false"
-
-	return extension
+func FailingReleaseExtension() plugins.ReleaseExtension {
```
drshrey: How is this a `FailingReleaseExtension` if it's just returning a `BasicReleaseExtensionService`?  If there's no difference, then why create a func for `FailingReleaseExtension`?

plugins/kubernetes/deployment_test.go
```diff
@@ -146,38 +167,47 @@ func (suite *TestSuiteDeployment) TestFailedDeployUnwindFirstDeploy() {
...
+			}
 
+			clientset.Extensions().Deployments(namespace).Update(&res.Items[0])
+
+			hasFailedDeployment = true
+		}
+
+		e, err = suite.transistor.GetTestEvent(plugins.GetEventName("release:kubernetes:deployment"), transistor.GetAction("status"), 30)
 		if err != nil {
 			assert.Nil(suite.T(), err, err.Error())
```
drshrey: I'd recommend using `assert.FailNow`. Then, you don't need to use the return statement afterwards either (https://godoc.org/github.com/stretchr/testify/assert#FailNow)

plugins/kubernetes/deployment.go
```diff
@@ -1384,98 +1394,163 @@ func (x *Kubernetes) unwindFailedDeployment(clientset kubernetes.Interface, name
...
 
-					log.Warn(fmt.Sprintf("SCALING DOWN REPLICASET FROM %d REPLICAS", rs.Status.Replicas))
-					_, err = x.ExtReplicaSetter.UpdateScale(clientset, namespace, rs.Name, scale)
-					if err != nil {
-						couldNotScale = true
-						log.Error(err)
-					}
-				}
+	// This is used in case the above updatescale operation does not complete.
+	// In that case we'll need to delete all the pods that are in this namespace
```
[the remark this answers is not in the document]
aballman: I haven't looked through the source for kubectl, but in practice if you delete the replica set via the API it does NOT delete the associated pods. I tell the RS to scale down here to initiate the termination process and then delete the RS. If that all fails, then I just delete the pods.

plugins/kubernetes/deployment.go
aballman: Same answer as your other question

plugins/kubernetes/deployment.go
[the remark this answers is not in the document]
drshrey: I see! I did find the `PropagationPolicy` option in DeleteOptions that seems relevant to this problem: https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#DeleteOptions. If we specify `Foreground`, then it will delete dependents of that particular object.

I'm not sure if we should handle dependent deletion ourselves if they already provide a way to do it. More information can be found here: https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#additional-note-on-deployments.

plugins/kubernetes/deployment_test.go
aballman: I'm aware of assert.FailNow. This works just fine. I'm not going to go and update all the cases of assert.Nil return for all of the old tests.

plugins/kubernetes/deployment.go
```diff
@@ -1384,141 +1396,178 @@ func (x *Kubernetes) unwindFailedDeployment(clientset kubernetes.Interface, name
...
-						},
-						ObjectMeta: meta_v1.ObjectMeta{
-							Name:      rs.Name,
-							Namespace: namespace,
-						},
-						Spec: v1beta1.ScaleSpec{
-							Replicas: 0,
-						},
-					}
+	if firstDeploy {
```
amoskyler: kind of personal preference I suppose, but I think you can collapse this statement into the if/else on line 1399

PR #430, opened 2019-02-11
Kong ingress deps

amoskyler:
- Removes Influxdbdata packages
- Adds gokong SDK packages
- Pins development postgres to 10.x


configs/circuit.yml
```diff
@@ -20,12 +20,13 @@ plugins:
...
   gitsync:
-    workers: 1
+    workers: 0
     workdir: "/tmp/gitsync" 
   heartbeat:
     workers: 0
   kubernetes:
-    workers: 1  
+    workers: 1
+    kong_api_url: http://192.168.39.175:31338
```
drshrey: Is the `kong_api_url` correct? This seems like a real IP where instead we should a placeholder format e.g. `http://<ip>:<port>`. Thoughts on this approach?

configs/circuit.yml
```diff
@@ -20,12 +20,13 @@ plugins:
       password: ""
     service_address: ":3011"
   gitsync:
-    workers: 1
+    workers: 0
```
drshrey: Is there a reason for 0? If not, we should revert this back to 0.

plugins/kubernetes/ingress_kong.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	projectSlug := plugins.GetSlug(payload.Project.Repository)
+
+	coreInterface := clientset.Core()
+	deploymentName := x.GenDeploymentName(projectSlug, inputs.Service.Name)
+
+	// var servicePorts []v1.ServicePort
```
drshrey: Should this be removed? If not, let's add a message explaining why this is commented out.

plugins/kubernetes/ingress_kong.go
```diff
@@ -0,0 +1,265 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: inputs.Service.ID,
+		},
+		Spec: serviceSpec,
+	}
+
+	service := coreInterface.Services(namespace)
+	svc, err := service.Get(inputs.Service.ID, metav1.GetOptions{})
+	var serviceObj *v1.Service
+	switch {
```
drshrey: Nice use of switch 👍

plugins/kubernetes/ingress_kong.go
```diff
@@ -0,0 +1,265 @@
...
+					}
+				}
+			}
+		}
+		serviceParams.ObjectMeta.ResourceVersion = svc.ObjectMeta.ResourceVersion
+		serviceParams.Spec.ClusterIP = svc.Spec.ClusterIP
+		serviceObj, err = service.Update(&serviceParams)
+		if err != nil {
+			return nil, fmt.Errorf("Error: failed to update service: %s", err.Error())
+		}
```
drshrey: nit: newline after `if` statement

plugins/kubernetes/ingress_kong.go
```diff
@@ -0,0 +1,265 @@
...
+			return nil, err
+		}
+
+		// Guarantee persisted ingress controller is configured on the extension side.
+		found := false
+		for _, controller := range strings.Split(ingressControllers.String(), ",") {
+			if controller == selectedIngress.String() {
+				found = true
+			}
+			continue
```
drshrey: remove unnecessary `continue`

plugins/kubernetes/ingress_kong.go
```diff
@@ -0,0 +1,265 @@
...
+
+	}
+
+	return &input, nil
+
+}
+
+func parseUpstreamRoutes(a transistor.Artifact) []UpstreamRoute {
+	var upstreamRoutes []UpstreamRoute
+	for _, upstream := range a.Value.([]interface{}) {
```
drshrey: Can we check that the type assertions is valid beforehand? Otherwise, we'll get a silent failure and execution will also stop. You can do so like this:
```
upstreamRoutes, ok := a.Value.([]interface{})
if !ok {
  log.Debug(...)
}
```

More info can be found here: https://stackoverflow.com/questions/37609749/how-to-log-type-assertion-errors-in-golang

plugins/kubernetes/ingress_kong.go
```diff
@@ -0,0 +1,265 @@
...
+
+		if err != nil {
+			log.Error(err)
+			x.sendErrorResponse(e, err.Error())
+		}
+	}
+}
+
+func (x *Kubernetes) deleteKongIngress(e transistor.Event) error {
+	return nil
```
drshrey: should this be empty?

plugins/kubernetes/ingress_kong.go
```diff
@@ -0,0 +1,265 @@
...
+		return err
+	}
+
+	log.Error(input)
+
+	return nil
+}
+
+func (x *Kubernetes) updateKongIngress(e transistor.Event) error {
+	return nil
```
drshrey: should this be empty?

amoskyler:
I messed this PR up with extra commits which I meant to reserve for later. Closing and will open a new PR with cherry picked commits.

PR #446, opened 2019-03-29
Add release env

zhouzhuojie:
Add `KUBE_POD_RELEASE_ENV`, I'm not sure if `FieldPath: "metadata.labels.environment"` can get the environment or not.

### DIFF
--- a/plugins/kubernetes/deployment.go
+++ b/plugins/kubernetes/deployment.go
@@ -16,7 +16,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	apis_batch_v1 "k8s.io/api/batch/v1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/api/extensions/v1beta1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/deployment.go
```diff
@@ -1330,5 +1330,14 @@ func (x *Kubernetes) exposePodInfoViaEnvVariable(myEnvVars []v1.EnvVar) []v1.Env
...
 
+	myEnvVars = append(myEnvVars, v1.EnvVar{
+		Name: "KUBE_POD_RELEASE_ENV",
+		ValueFrom: &v1.EnvVarSource{
+			FieldRef: &v1.ObjectFieldSelector{
+				FieldPath: "metadata.labels.environment",
+			},
+		},
+	})
+
```
ziru: release env, project, service is a codeamp abstraction, instead of k8s abstraction.   better name it CODEAMP_RELEASE_ENV.

also, codeamp passes such release info to this plugin via `reData`, and can be accessed via `reData.Release.Environment`

amoskyler:
This looks alright to me. Can I get a second confirm on that @drshrey?

plugins/kubernetes/deployment.go
zhouzhuojie: updated

zhouzhuojie:
image


@drshrey do you know why circleci is pending?

aballman:
> image
> 
> @drshrey do you know why circleci is pending?

@zhouzhuojie Its because we didn't previously enable circleci to build forked branches. I've enabled that now so if you push a noop it should run.

zhouzhuojie:
@aballman 

```
No CODECLIMATE_REPO_TOKEN found. A CODECLIMATE_REPO_TOKEN must be specified as an environment variable.
Exited with code 1
```

CircleCI for public forks cannot access some private env variables.

zhouzhuojie:
> @aballman
> 
> ```
> No CODECLIMATE_REPO_TOKEN found. A CODECLIMATE_REPO_TOKEN must be specified as an environment variable.
> Exited with code 1
> ```
> CircleCI for public forks cannot access some private env variables.

Since it's already open source, and code climate supports it, https://codeclimate.com/oss/, do we need a private token here?

PR #447, opened 2019-04-03
Private Loadbalancers Whitelist

aballman:
* Allows Private/Office type loadbalancers to augment their whitelisted IP ranges


drshrey:
@aballman looks good, can we add tests for this?

aballman:
> @aballman looks good, can we add tests for this?

@drshrey No, not really. I've added the artifact here just to increase the test coverage so it's not decreasing the coverage. But testing load balancers / services isn't totally possible.

PR #448, opened 2019-04-04
Allow Project Extension to Append Whitelist Rules for ELB

aballman:
* Allow the Project Extension LB configuration to specify additional whitelist rules for external traffic

### DIFF
--- a/plugins/kubernetes/loadbalancer.go
+++ b/plugins/kubernetes/loadbalancer.go
@@ -139,7 +139,6 @@ func (x *Kubernetes) doLoadBalancer(e transistor.Event) error {
 	*
 	*********************************************/
 	var loadBalancerSourceRanges []string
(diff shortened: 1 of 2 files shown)


plugins/kubernetes/loadbalancer.go
```diff
@@ -138,7 +138,16 @@ func (x *Kubernetes) doLoadBalancer(e transistor.Event) error {
 	*	Prepare Annotations and ServiceType
 	*
 	*********************************************/
-	var loadBalancerSourceRanges []string
+	loadBalancerSourceRanges := make([]string, 0, 10)
+
+	projectLBSourceRanges, err := e.GetArtifact("projext_lb_source_ranges")
```
sasso: Typo `projext_...`?

aballman: short for "project extension" because we don't have a good way of keeping artifact names from colliding...

PR #454, opened 2019-04-25
S3 Self Service Plugin

aballman:
* Allows a user to provision an IAM identity that allows access to a specified path in a bucket
* Deleting the extension revokes access to the data, but retains the data to be dealt with manually
* Provides credentials for users to ferry over to the secrets page and store as env variable
https://checkr.atlassian.net/browse/EDEV-1157

### DIFF
--- a/plugins/s3/interface.go
+++ b/plugins/s3/interface.go
@@ -1,0 +1,60 @@
+package s3
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/aws/aws-sdk-go/service/s3"
+)
+
+type S3API interface {
+	PutObject(*s3.PutObjectInput) (*s3.PutObjectOutput, error)
+}
+
+type IAMAPI interface {
+	AddUserToGroup(*iam.AddUserToGroupInput) (*iam.AddUserToGroupOutput, error)
+	RemoveUserFromGroup(*iam.RemoveUserFromGroupInput) (*iam.RemoveUserFromGroupOutput, error)
+
+	CreateAccessKey(*iam.CreateAccessKeyInput) (*iam.CreateAccessKeyOutput, error)
+	ListAccessKeys(*iam.ListAccessKeysInput) (*iam.ListAccessKeysOutput, error)
+	DeleteAccessKey(*iam.DeleteAccessKeyInput) (*iam.DeleteAccessKeyOutput, error)
+
+	CreateUser(*iam.CreateUserInput) (*iam.CreateUserOutput, error)
+	GetUser(*iam.GetUserInput) (*iam.GetUserOutput, error)
+	DeleteUser(*iam.DeleteUserInput) (*iam.DeleteUserOutput, error)
+	WaitUntilUserExists(*iam.GetUserInput) error
+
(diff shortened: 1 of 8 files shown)


plugins/s3/s3.go
```diff
@@ -0,0 +1,364 @@
...
+		"project:s3:create",
+		"project:s3:update",
+		"project:s3:delete",
+	}
+}
+
+// How does this work?
+// A bucket is shared by other staging projects
+// in order to create access for a new project
+// to use htis storage, we will need to generate
```
amoskyler: "htis" is typo

plugins/s3/s3.go
```diff
@@ -0,0 +1,364 @@
...
+                "arn:aws:s3:::%s/%s/*"
+            ]
+        }
+    ]
+	}`
+
+	userPolicy := fmt.Sprintf(userPolicyTemplate, x.awsBucket, payload.Project.Slug)
+	_, err = iamSvc.PutUserPolicy(&iam.PutUserPolicyInput{
+		UserName:       &userName,
+		PolicyName:     aws.String("codeamp-storage"),
```
amoskyler: So I am not entirly familiar with the AWS api, but is this policy independent of the IAM policy resource? Is it effectively an inline policy? I am attempting to understand if there is a possiblity of name collisions with the policy-name.

plugins/s3/s3.go
```diff
@@ -0,0 +1,364 @@
...
+	// Generate an API key for the user
+	accessKeyResponse, err := iamSvc.CreateAccessKey(&iam.CreateAccessKeyInput{
+		UserName: &userName,
+	})
+	if err != nil {
+		log.Error(err.Error())
+		return err
+	}
+
+	secondsToWait := 10
```
amoskyler: Is this a sufficiently long wait time? What happens if an error occurs in the following code (Or the wait time isn't tolerant enough)?

It looks like the update logic lower down doesn't go through any of this procedure, so if some error occurs in here, will it be recoverable?

My current intuition is if there is a failure, the user can delete the extension (wihch will delete the user we created just above), allowing them the ability to re-install the extension.

plugins/s3/s3.go
aballman: That is exactly what it is. User policies only apply for that specific user. We tend to not use them often because they are limited in size per policy per user. That limit is usually much lower than a regular policy or even a group policy.

plugins/s3/s3.go
aballman: I've updated the above to be a configurable timeout from the extensions artifacts.
I've set the interval to 5s and the extension configuration is 100. Hopefully should be enough! Really just need a few moments for S3 to realize a new key has been made for it.

plugins/s3/s3_test.go
```diff
@@ -0,0 +1,237 @@
...
+		if e.State != "running" {
+			break
+		}
+	}
+
+	suite.T().Log(e.StateMessage)
+	assert.Equal(suite.T(), transistor.GetState("failed"), e.State)
+}
+
+// func (suite *TestSuiteS3Extension) TestDeleteS3ExtSuccess() {
```
amoskyler: Plans to complete these eventually? Or should they be deleted?

plugins/s3/interface.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/aws/aws-sdk-go/service/s3"
+)
+
+type S3API interface {
+	PutObject(*s3.PutObjectInput) (*s3.PutObjectOutput, error)
+}
+
+type IAMAPI interface {
```
amoskyler: Nice interfaces.

Issue #457, opened 2019-04-29
Fix db password generator


drshrey:
Currently, the database password generator function will output the same sequence given the same time value as the input in the `Seed` function. If the system clock was hacked to be the same value every time this function was called, then passwords would be the same for all dbs created afterwards. We should make it so the sequence is random no matter what every single time the function is invoked.

PR #458, opened 2019-04-30
Add mysql plugin 

drshrey:
also fixes https://github.com/codeamp/circuit/issues/457


plugins/database/database.go
```diff
@@ -124,9 +124,13 @@ func (x *Database) Process(e transistor.Event) error {
 	case transistor.GetAction("create"):
```
amoskyler: What happens if a user hits the "update" button?

plugins/database/database.go
```diff
@@ -141,6 +146,8 @@ func (x *Database) Process(e transistor.Event) error {
 		}
 
 		x.sendSuccessResponse(e, transistor.GetState("complete"), artifacts)
+	case transistor.GetAction("update"):
+		x.sendSuccessResponse(e, transistor.GetState("complete"), e.Artifacts)
```
amoskyler: I believe `e.Artifacts` will include all artifacts, including other extensions artifacts. I think we need to pick out the specific artifacts we want to return to the user (only the artifacts which are relevant to this extension).

PR #459, opened 2019-04-30
Add metrics endpoint

sasso:
Collect metrics from Postgres and Redis


plugins/codeamp/metrics/collector/collector.go
```diff
@@ -0,0 +1,174 @@
...
+func NewCollector(redisOpts RedisCollectorOpts, postgresOpts PostgresCollectorOpts) *Collector {
+	exporter := &Collector{
+		RedisOpts:    redisOpts,
+		PostgresOpts: postgresOpts,
+	}
+
+	return exporter
+}
+
+func (exporter *Collector) Describe(ch chan<- *prometheus.Desc) {
```
amoskyler: What is the relationship between an `exporter` and a `*Collector`? Why isn't `exporter` just called `*Collector` or better yet, `c`?

sasso: will fix!

plugins/codeamp/metrics/collector/collector.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func (c *Collector) collectProjectSettings(db *gorm.DB) {
+	var projects []model.Project
+	var environments []model.Environment
+
+	db.Find(&projects)
+	db.Find(&environments)
+
+	for _, project := range projects {
```
amoskyler: Quick note:
We could increase the performance of this collector significantly by iterating over environments and then querying projects joined with project settings.

Something like

```
db.Find(&environments)
for _, environment := range environments {
  // select * from projects join projects.id = project_settings.project_id where environment_id = environment.id
}
```
This would allow us to perform 1 + N queries, where N is the environment count. As opposed to 2 + N queries where N is total number of projects.

PR #462, opened 2019-05-01
Patch: change genDBUser func to have 14 chars

drshrey:
Change the genDBUser func to have 10 chars and only use the uuid chars. 

Improvement: A longer term fix is to allow variable lengths depending on the DB extension's artifact and use that as input when generating the function.


drshrey:
Currently, we cannot install mysql DBs because the max username character length for mysql is 16 characters.

PR #467, opened 2019-05-02
Mongo Self Service Plugin

aballman:
* Allow users to create their own Mongo resources


aballman:
Will remove 371f27150d78a38b9cc756574e51837effbd7ac9 and e9a01782a2c6458f09b9c8194aac8a14f54f5f24
when #468 and #469 are merged.

PR #474, opened 2019-05-15
Add Complainer logic for notifying when features have been successfully deployed to "production" before "staging"

drshrey:


plugins/codeamp/release_handler.go
```diff
@@ -177,7 +177,17 @@ func (x *CodeAmp) ReleaseCompleted(release *model.Release) {
...
-		x.SendNotifications("SUCCESS", release, &project)
+		complained, err := x.ComplainIfNotInStaging(release, &project)
+		if err != nil {
+			log.InfoWithFields(err.Error(), log.Fields{
+				"release": release,
+			})
+			return
+		}
+
+		if !complained {
```
amoskyler: What is the purpose of this if block?

plugins/codeamp/constants/constants.go
```diff
@@ -0,0 +1,6 @@
+package constants
+
+import "os"
+
+var StagingEnvironment = os.Getenv("STAGING_ENVIRONMENT")
```
amoskyler: Why is this in the "constants" package? This is the opposite of a constant.

plugins/codeamp/release_handler.go
drshrey: If the complained notification is sent out, I don't want to also send out the SUCCESS message since that is redundant (they're both sent to the same channel).

plugins/codeamp/constants/constants.go
drshrey: So it is a constant but we need to store it as a var because of the constraints of Go. We cannot use constants for values that are passed during runtime i.e. environment variables in this case.

From this article - https://medium.com/rungo/variables-and-constants-in-go-programming-c715443fa788
```
In Go, constant value must be known during compile time. If you are trying to assign value returned by a function call, that won’t work. const a = math.Sqrt(4) is an illegal statement as value of a is computed during run time.
```

plugins/codeamp/complainer.go
```diff
@@ -2,15 +2,20 @@ package codeamp
...
 	"github.com/jinzhu/gorm"
 
-	"github.com/codeamp/circuit/plugins/codeamp/constants"
 	"github.com/codeamp/circuit/plugins/codeamp/model"
 	log "github.com/codeamp/logger"
 	"github.com/codeamp/transistor"
 )
 
+var (
+	PARENT_ENVIRONMENT = os.Getenv("PARENT_ENVIRONMENT")
```
amoskyler: These should be set in the codeamp config file. There should also be an option to disable complainer.

PR #475, opened 2019-05-17
Scheduled Build Releaser

aballman:
* Adds plugin to enable a scheduled build on a designated branch.
* User can specify which branch and what time to initiate a release
* Can be set for global or project specific label
* Will update the project and notify the front end if the extension is attached to a project with a branch that's not set to the 'designated' branch. If that's true, and its the scheduled time, it will update branch and kickoff a release.


plugins/codeamp/workflowreleaseextension_handler.go
```diff
@@ -49,13 +48,22 @@ func (x *CodeAmp) WorkflowReleaseExtensionsCompleted(release *model.Release) {
 
 	user := model.User{}
 	email := ""
-	if release.UserID == uuid.FromStringOrNil(ContinuousDeployUUID) {
-		email = "Automated Deployment"
+
+	specialReleaseUsers := map[string]string{
+		ContinuousDeployUUID: "Automated Deployment",
```
amoskyler: does `ContinuousDeployUUID` resolve to the string value `59ee0229-9814-4d9b-be83-7921fe6069c1` in this context? Or is it the literal key `ContinuousDeployUUID`?

plugins/codeamp/workflowreleaseextension_handler.go
```diff
@@ -49,13 +48,22 @@ func (x *CodeAmp) WorkflowReleaseExtensionsCompleted(release *model.Release) {
...
+		ScheduledDeployUUID:  "Scheduled Deployment",
+	}
+	if val, ok := specialReleaseUsers[release.UserID.String()]; ok {
+		email = val
 	} else {
-		if x.DB.Where("id = ?", release.UserID).First(&user).RecordNotFound() {
-			log.InfoWithFields("user not found", log.Fields{
-				"id": release.UserID,
-			})
+		if err := x.DB.Where("id = ?", release.UserID).First(&user).Error; err != nil {
```
amoskyler: Do we manually add this user to the user's table?

plugins/codeamp/workflowreleaseextension_handler.go
```diff
@@ -49,13 +48,22 @@ func (x *CodeAmp) WorkflowReleaseExtensionsCompleted(release *model.Release) {
...
+			if gorm.IsRecordNotFoundError(err) {
```
amoskyler: We don't bail out of this if the user is invalid?

plugins/scheduledbranchreleaser/scheduledbranchreleaser.go
```diff
@@ -0,0 +1,155 @@
...
+		log.InfoWithFields(fmt.Sprintf("Process ScheduledBranchReleaser event: %s", e.Event()), log.Fields{})
+
+		payload := e.Payload.(plugins.ScheduledBranchReleaser)
+		timeScheduledToBuild, err := e.GetArtifact("schedule")
+		if err != nil {
+			log.Error(err.Error())
+			return err
+		}
+
+		t, err := time.Parse("15:04 -0700 MST", timeScheduledToBuild.String())
```
amoskyler: Nit: timezone should be an application level configuration

plugins/scheduledbranchreleaser/scheduledbranchreleaser.go
```diff
@@ -0,0 +1,155 @@
...
+		if err != nil {
+			log.Error(err.Error())
+			return err
+		}
+
+		// Grab the current time and date, then truncate the time so we're left with only today's date
+		// Then add the time that we have from the SCHEDULE so we can calculate the duration between
+		// now and the scheduled time for today
+		now := time.Now().UTC()
```
amoskyler: I just want to make sure I get the gist of this:

We parse `timeScheduledToBuild.String()` and assume the time entered is in `MST` (up on L#116), and convert to UTC. Then we get the current time and check the difference between the current time and the scheduled time?

plugins/codeamp/codeamp.go
```diff
@@ -198,9 +201,10 @@ func (x *CodeAmp) Subscribe() []string {
 	return []string{
 		"gitsync:status",
 		"heartbeat",
+		"scheduledbranchreleaser:release",
 		"websocket",
-		"project",
-		"release",
+		"^project",
```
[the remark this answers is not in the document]
aballman: This is a regex, so the `^` is an anchor to match only strings where this is the very beginning of the text. It was matching my "scheduledbranchreleaser" messages because it has `release` in it so i changed it to `^release`

plugins/scheduledbranchreleaser/scheduledbranchreleaser.go
aballman: what? this is the sprintf format that go uses. it is not the actual timezone, thats provided via project extension configuration

amoskyler:
Understood.

plugins/codeamp/workflowreleaseextension_handler.go
aballman: It resolves to the string value in this context. Basically it's a map to convert the special user ids into a readable informative string. If the userID key doesn't exist in that map, then we query the database to see if it's a "real user id"

plugins/codeamp/workflowreleaseextension_handler.go
aballman: I'm not sure what you're asking here? I updated this block to distinguish `RecordNotFound` errors from other DB errors that might involve connectivity or missing fields, etc

plugins/codeamp/workflowreleaseextension_handler.go
aballman: Not according to the history, which isn't entirely crazy. The user query there is so we can pass it downstream to the notification plugins, it's not really needed for wrapping up the workflow release extension here.

plugins/codeamp/codeamp.go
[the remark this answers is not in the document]
drshrey: Was the anchor added for the same reason for `project`? If so, should we add anchors to all other subscribed events as well e.g. websocket, heartbeat, gitsync?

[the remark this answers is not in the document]
aballman: Yes. Only if there is going to be overlap, which in this case there was. We can update those but I'd prefer to do that in another PR.

plugins/codeamp/scheduledbranchreleaser_handler.go
```diff
@@ -0,0 +1,441 @@
...
+
+// This has to be here because we need a way of referencing the extension
+// in order to get the extension ID and then grab all the project extensions
+// that have that extension id as their parent extension
+const (
+	SCHEDULED_RELEASE_HANDLER_PLUGIN_NAME = "scheduledbranchreleaser"
+)
+
+// This is the handler that is automagically called when
+// CodeAmp receives a message with a 'ScheduleBranchReleaser' type payload
```
drshrey: nit: Schedule**d**BranchReleaser

plugins/codeamp/scheduledbranchreleaser_handler.go
```diff
@@ -0,0 +1,441 @@
...
+	if err != nil {
+		log.Error(err.Error())
+		return
+	}
+
+	// Find the project settings for the project in question
+	var projectSettings model.ProjectSettings
+	if err := x.DB.Where("id = ?", payload.ProjectSettingsID).Find(&projectSettings).Error; err != nil {
+		log.Error(err.Error())
+		return
```
drshrey: no need for return here since we have `else`

plugins/codeamp/scheduledbranchreleaser_handler.go
```diff
@@ -0,0 +1,441 @@
...
+	} else {
+		if projectSettings.GitBranch != desiredBranch.String() {
+			oldBranch := projectSettings.GitBranch
+
+			// Try and find the envrionment that is associted with the environment id that was provided
```
drshrey: nit: environment* associated*

plugins/codeamp/scheduledbranchreleaser_handler.go
```diff
@@ -0,0 +1,441 @@
...
+			// in the message. this is necessary so we can send a message to the front end to inform
+			// the user that the branch has been updated without their explicit input
+			var environment model.Environment
+			if err := x.DB.Where("id = ?", payload.Environment).Find(&environment).Error; err != nil {
+				log.Error(err.Error())
```
drshrey:
```
log.ErrorWithFields(err.Error(), log.Fields{
   "id": payload.Environment
}
```

Also, it should be called `EnvironmentID`. Also, do we have access to `Environment` in this object?  Based off the definition of `plugins.ScheduledBranchReleaser`, it doesn't look like it: 
```
type ScheduledBranchReleaser struct {
	ProjectExtension  `json:"projectextension"`
	Git               `json:"git"`
	ProjectSettingsID string `json:"projectSettingsID"`
}
```
https://github.com/codeamp/circuit/pull/475/files#diff-17398cb2b8985e353f15405e8d9b4274R241

plugins/codeamp/scheduledbranchreleaser_handler.go
```diff
@@ -0,0 +1,441 @@
...
+				return
+			}
+
+			// Update the project settings DB entry to reflect the new branch selection
+			{
```
drshrey: What're the brackets for? I've never seen this before, so mostly just curious

plugins/codeamp/workflowreleaseextension_handler.go
amoskyler: You are right, I missed that.

plugins/codeamp/workflowreleaseextension_handler.go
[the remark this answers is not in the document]
aballman: because i want to be able to distinguish in the logs if it wasn't found or if it was found but deleted

plugins/codeamp/scheduledbranchreleaser_handler.go
aballman: i like being explicit with the returns here in case someone comes and edits the code and adds something after the else statement.

plugins/codeamp/scheduledbranchreleaser_handler.go
aballman: these aren't that helpful

plugins/codeamp/scheduledbranchreleaser_handler.go
aballman: organization

plugins/codeamp/scheduledbranchreleaser_handler.go
aballman: Then someone needs to update the ProjectExtension struct and everywhere it is used.
When I use payload.Environment its shorthand for payload.ProjectExtension.Environment.

plugins/codeamp/projectextension_handler.go
```diff
@@ -17,11 +17,16 @@ func (x *CodeAmp) ProjectExtensionEventHandler(e transistor.Event) error {
...
-		if x.DB.Where("id = ?", receivedPayload.ID).Find(&extension).RecordNotFound() {
+		if x.DB.Unscoped().Where("id = ?", receivedPayload.ID).Find(&extension).RecordNotFound() {
 			log.ErrorWithFields("extension not found", log.Fields{
 				"id": receivedPayload.ID,
 			})
+
 			return fmt.Errorf(fmt.Sprintf("Could not handle ProjectExtension status event because ProjectExtension not found given payload id: %s.", receivedPayload.ID))
+		} else {
+			if extension.DeletedAt != nil {
+				return fmt.Errorf(fmt.Sprintf("Could not handle ProjectExtension status event because ProjectExtension has been deleted id: %s.", receivedPayload.ID))
```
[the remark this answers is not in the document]
aballman: that's a good point, but i didn't write this code. you should open an issue to update this in other places in the codebase as well, and i'll fix the cases in this file.

PR #477, opened 2019-05-22
S3 Plugin Creates Buckets

aballman:
Previously the S3 plugin used a pre-allocated bucket and then created permissions based on prefixes to slice up the bucket for other projects to share. This is inappropriate for some projects and they require a full bucket, such as in the case of ActiveRecord with its default settings. These changes modify the plugins behavior to create new buckets.

### DIFF
--- a/plugins/s3/s3.go
+++ b/plugins/s3/s3.go
@@ -21,7 +21,7 @@ import (
 
 var src = rand.NewSource(time.Now().UnixNano())
 
-const letterBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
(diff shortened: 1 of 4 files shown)


plugins/s3/s3.go
```diff
@@ -18,6 +19,33 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3"
 )
 
+var src = rand.NewSource(time.Now().UnixNano())
+
+const letterBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
```
amoskyler: Can you get rid of the capital letters? It will make the bucket names unaesthetic and gross.

PR #480, opened 2019-06-05
kong ingress extension input handling

amoskyler:
1. Trim whitespace from http method inputs
2. allow case sensitive paths

### DIFF
--- a/plugins/kubernetes/ingress_kong.go
+++ b/plugins/kubernetes/ingress_kong.go
@@ -586,7 +586,7 @@ func parseUpstreamRoutes(a transistor.Artifact) ([]UpstreamRoute, error) {
 		if pathsString == "" {
 			paths = []string{}
 		} else {
-			paths = strings.Split(strings.ToLower(pathsString), ",")
+			paths = strings.Split(pathsString, ",")
 		}
 
 		domains := upstream.(map[string]interface{})["domains"]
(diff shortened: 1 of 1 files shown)


plugins/kubernetes/ingress_kong.go
```diff
@@ -229,6 +229,9 @@ func (x *Kubernetes) createKongIngress(e transistor.Event) error {
 
 		if len(route.Methods) > 0 {
 			methods := strings.Split(strings.ToUpper(strings.Join(route.Methods, ",")), ",")
```
drshrey: Could also do a `strings.Replace(..., " ", "", -1)`, but I'm fine with approving this for now

amoskyler: Why would `strings.Replace()` be preferred in this situation?

drshrey: We may also want to return an error if we find spaces in indexes that are not at the front or end of the string e.g. `route1, route    2, route3`

amoskyler: I think that's unnecessary in this case. Kong will respond with a failure directly. This is mostly for user convienience. If the user is attempting to create a method of `POST ABC` then it will be validated by the kong api.

This is mostly to help users who add spaces when making their list `POST, GET`, and it's not totally clear when the API responds with validation that it's literally attempting to add ` GET`

PR #483, opened 2019-06-11
Kong ingress extension tests

amoskyler:


plugins/kubernetes/ingress_kong.go
```diff
@@ -85,17 +91,21 @@ func (x *Kubernetes) deleteKongIngress(e transistor.Event) error {
 		existingKongService, err := kongClient.Services().GetServiceByName(inputs.Service.ID)
 		if err != nil {
 			log.Error(err)
+			x.sendErrorResponse(e, "failed to delete service")
```
aballman: theoretically could send extra success message if an `err` was ever returned in conjunction with a nil `existingKongService` as it would send the error then the success message

plugins/kubernetes/ingress_kong.go
```diff
@@ -57,17 +59,20 @@ func (x *Kubernetes) deleteKongIngress(e transistor.Event) error {
...
 		log.Error(err)
 		log.Error(status)
+		x.sendErrorResponse(e, "failed to delete service")
 		return err
 	}
 
 	ingType, err := e.GetArtifact("type")
 	if err != nil {
 		return fmt.Errorf("Failed to retrieve service type")
+		x.sendErrorResponse(e, "failed to delete service")
```
aballman: this line will never be hit

plugins/kubernetes/ingress_helpers.go
```diff
@@ -58,7 +57,7 @@ func getTableViewFromDomains(domains []Domain) string {
 }
 
 // Service should be in the format servicename:port
-func parseService(e transistor.Event) (Service, error) {
+func ParseService(e transistor.Event) (Service, error) {
```
aballman: why did you export this?

plugins/kubernetes/ingress_kong.go
amoskyler: Good point.

plugins/kubernetes/ingress_helpers.go
amoskyler: In order to call `ParseService(e)` from the test. Purpose being to verify the service name in kong.

plugins/kubernetes/ingress_kong.go
amoskyler: fixed.

PR #487, opened 2019-06-26
fix #301 in panel, but only for public repos

davechan813:
- Add a get request to github api to check if the public repo exists or not. If not, return error message to frontend and prevent creation.

### DIFF
--- a/plugins/codeamp/graphql/project_mutation.go
+++ b/plugins/codeamp/graphql/project_mutation.go
@@ -8,6 +8,7 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"net/http"
 	"strings"
 
 	"github.com/codeamp/circuit/plugins"
@@ -27,6 +28,8 @@ import (
 type ProjectResolverMutation struct {
 	DB *gorm.DB
 }
+
+const GITHUB_REPO_API string = "https://api.github.com/repos"
 
 // CreateProject Create project
 func (r *ProjectResolverMutation) CreateProject(ctx context.Context, args *struct {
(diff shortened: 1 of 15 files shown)


plugins/codeamp/graphql/project_mutation.go
```diff
@@ -28,6 +29,8 @@ type ProjectResolverMutation struct {
 	DB *gorm.DB
 }
 
+const GITHUB_REPO_API string = "https://api.github.com/repos"
```
amoskyler: Unfortunately this is too limited a scope for this feature. Even though we don't use this in our infra, you can add git projects from any valid git repository (not just ones hosted on github).

amoskyler: This breaks functionality where the repo uri is not hosted on github.com

plugins/codeamp/graphql/project_mutation.go
```diff
@@ -60,6 +63,14 @@ func (r *ProjectResolverMutation) CreateProject(ctx context.Context, args *struc
 		return nil, fmt.Errorf("This repository already exists. Try again with a different git url.")
 	}
 
+	// Check if project exists in github
+	if protocol == "HTTPS" { // now only check for public repo
+		resp, err := http.Get(fmt.Sprintf("%s/%s/%s", GITHUB_REPO_API, res["owner"], res["repo"]))
```
amoskyler: To address my previous comments, instead of using the new constant `GITHUB_REPO_API` you can either

1. use `args.Project.GitUrl`as the http get path. Note: if this is done redirects will have to be followed (and the last status code tested) as github will 301 redirect all GET requests on`/repo/project.git` paths to `/repo/project`. After the redirect the value should either be 200 or 404 depending on the existence of the project.

2. Use the **host**, owner, and repo values from the regex params.

davechan813: Good catch! I adopted the first way and it turned out it was an easy fix. Golang follows redirection automatically so I just changed the link to `args.Project.GitUrl`. I tested for several repos hosted on Github, Gitlab, Gitbucket, Bitbucket, and they work. I also changed the error message to show which site the repo is hosted.

Screen Shot 2019-07-01 at 10 54 35 AM
Screen Shot 2019-07-01 at 10 56 49 AM

plugins/codeamp/graphql/project_mutation.go
```diff
@@ -65,9 +65,9 @@ func (r *ProjectResolverMutation) CreateProject(ctx context.Context, args *struc
 
 	// Check if project exists in github
 	if protocol == "HTTPS" { // now only check for public repo
-		resp, err := http.Get(fmt.Sprintf("%s/%s/%s", GITHUB_REPO_API, res["owner"], res["repo"]))
+		resp, err := http.Get(args.Project.GitUrl)
```
amoskyler: I'm relatively certain this will always return a 301 redirect. You must "follow redirects" and base your `resp.StatusCode != 200` at the end of the redirect.

PR #488, opened 2019-06-27
Add envsDeployedIn resolver for project

drshrey:
This defines a resolver within Project that allows a client to query which environments a feature has successfully been deployed within.


plugins/codeamp/db/project.go
```diff
@@ -152,6 +152,60 @@ func (r *ProjectResolver) ContinuousDeploy() bool {
 	}
 }
 
+// IsDeployed
```
sasso: Nit: this should be DeployedIn

plugins/codeamp/db/project.go
```diff
@@ -152,6 +152,60 @@ func (r *ProjectResolver) ContinuousDeploy() bool {
...
+	var environments []*EnvironmentResolver
+	var feature model.Feature
+	var allReleases []model.Release
+
+	// get feature
+	if err := r.DB.Where("project_id = ? and hash = ?", r.Project.Model.ID.String(), args.GitHash).Find(&feature).Error; err != nil {
+		return []*EnvironmentResolver{}, err
+	}
+
+	// get descendant features to check if they've been deployed too since that implies this one has been deployed too
```
sasso: We are checking parent, not descendant right?

plugins/codeamp/db/project.go
```diff
@@ -152,6 +152,60 @@ func (r *ProjectResolver) ContinuousDeploy() bool {
...
+		// get releases where head_feature_id matches this feature
+		if err := r.DB.Where("head_feature_id = ? and state = ?", tmpFeature.Model.ID.String(), transistor.GetState("complete")).Find(&releases).Error; err != nil {
+			return []*EnvironmentResolver{}, err
+		}
+
+		allReleases = append(releases)
+	}
+
+	// get unique environment_ids associated in list of releases
+	uniqueEnvironmentIDs := map[string]bool{}
```
sasso: you could do unique in the query above and ditch this loop.

plugins/codeamp/db/project.go
drshrey: We're checking descendant because if a feature that was created after the target feature was deployed successfully, then that descendant feature deployment includes the code of the target feature as well, implying the target feature code is "deployed" in the descendant features' deployed environments. This is why we check for descendants on top of just the target feature.

PR #492, opened 2019-07-16
Save branch setting when new environment added

### LINKED ISSUE #455: Provisioning new environment in CodeAmp
Current:
When I provision a new environment to a project in CodeAmp, deploys fail. In order to make them succeed, I need to re-save the branch setting. It looks like the branch setting is not being picked up by the newly provisioned environment.

Expected:
1. I provision a new environment
2. I click 'deploy'
3. Deploy recognizes branch and proceeds with deploy.

davechan813:
Fixed #455. Call `UpdateProject` with default setting when new environment provisioned.

### DIFF
--- a/plugins/codeamp/graphql/project_mutation.go
+++ b/plugins/codeamp/graphql/project_mutation.go
@@ -297,6 +297,30 @@ func (r *ProjectResolverMutation) UpdateProjectEnvironments(ctx context.Context,
 				EnvironmentID: environment.Model.ID,
 				ProjectID:     project.Model.ID,
 			}
+
+			// Save branch if this is a new environment
+			if r.DB.Where("environment_id = ? and project_id = ?", environment.Model.ID, project.Model.ID).Find(&projectEnvironment).RecordNotFound() {
+				projectID := fmt.Sprint(project.ID)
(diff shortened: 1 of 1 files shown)


plugins/codeamp/graphql/project_mutation.go
```diff
@@ -297,6 +297,36 @@ func (r *ProjectResolverMutation) UpdateProjectEnvironments(ctx context.Context,
 				EnvironmentID: environment.Model.ID,
 				ProjectID:     project.Model.ID,
 			}
+
+			// Save branch if this is a new environment
+			if r.DB.Where("environment_id = ? and project_id = ?", environment.Model.ID, project.Model.ID).Find(&projectEnvironment).RecordNotFound() {
```
aballman: Can you please use `gorm.IsRecordNotFoundError` (i think that's the right function)
to determine if it is a RecordNotFound error or not? If it's not a simple record not found, like there being a DB constraint or some sort of other error, we should log it.

You would do `if err := r.DB.Where(query).Find(&model).Error; err != nil` and then `if gorm.IsRecordNotFoundError(err)`

PR #493, opened 2019-07-16
Add labels to redis deployment

### LINKED ISSUE #490: Add labels to redis deployment to know which project is using it
When looking in kube, it's hard to cross-reference the actual project that has installed the redis deployment with the given name of the redis deployment. It would be great if we had labels to make it crystal clear which project is using which redis workers.

davechan813:
Fix #490. Sorry made a mistake pushing wrong branch last time.

### DIFF
--- a/plugins/kubernetes/redis.go
+++ b/plugins/kubernetes/redis.go
@@ -73,8 +73,8 @@ func (x *Kubernetes) doRedis(e transistor.Event) error {
 
 	payload := e.Payload.(plugins.ProjectExtension)
 
-	deploymentName := genRedisDeploymentName(payload)
-	redisDeployment, err := x.createRedisDeploymentSpec(deploymentName)
+	deploymentName, projectName := genRedisDeploymentName(payload)
+	redisDeployment, err := x.createRedisDeploymentSpec(deploymentName, projectName)
 	if err != nil {
 		x.sendErrorResponse(e, err.Error())
 		return err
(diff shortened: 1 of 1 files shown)


plugins/kubernetes/redis.go
```diff
@@ -204,20 +209,21 @@ func (x *Kubernetes) createRedisServiceSpec(payload plugins.ProjectExtension, sv
...
 	projectSlug := payload.Project.Slug
+	shortProjectSlug := payload.Project.Slug
 
 	if len(payload.Environment) > 10 {
 		env = payload.Environment[:10]
 	}
 
 	if len(payload.Project.Slug) > 10 {
-		projectSlug = payload.Project.Slug[:10]
+		shortProjectSlug = payload.Project.Slug[:10]
```
aballman: I think you should change this around a bit. Instead of keeping the data in two variables, projectSlug and shortProjectSlug, you should keep an int that stores the length of `projectSlug`.

Then, if the length is greater than the limit, you can set it to 10. Then use projectSlug for both fields.

```
projectSlugLength = max(10, len(projectSlug))
return fmt.Sprintf("%s-%s-%s", env, projectSlug[:projectSlugLength], uniqueID.String()[:8]), projectSlug
```

Otherwise everything else looks good

davechan813: Good idea, changed. Didn't end up using `math.Max` because it can only be used for `float` type.

PR #495, opened 2019-07-22
Handle forced pushed commits

### LINKED ISSUE #319: Handle Force Pushed Repository
**Describe the bug**
We keep track of releases based off of commits. When a user force pushes the project they are working on, it rewrites commits.

[] 1. Find a way to deal with commits that no longer exist
[] 2. Find a way to pull down changes in gitsync without choking on force pushed branch

davechan813:
Fix #319. Compare feature lists with the latest commits list pulled from Github. Add in `redeployable` and `redeployableMessage` fields. Add `notFoundSince` to record the time when the feature is not found.


plugins/gitsync/gitsync.go
```diff
@@ -220,9 +220,9 @@ func (x *GitSync) Process(e transistor.Event) error {
 
 			_commits = append(_commits, c)
 
-			if c.Hash == payload.From {
-				break
-			}
+			// if c.Hash == payload.From {
```
amoskyler: Why is this commented out? It should be deleted if it is not needed.

plugins/events.go
```diff
@@ -137,12 +137,13 @@ type GitSync struct {
...
-	User       string    `json:"user"`
-	Message    string    `json:"message"`
-	Created    time.Time `json:"created"`
+	ID            string    `json:"id"`
+	Hash          string    `json:"hash"`
+	ParentHash    string    `json:"parentHash"`
+	User          string    `json:"user"`
+	Message       string    `json:"message"`
+	Created       time.Time `json:"created"`
+	NotFoundSince time.Time `json:"notFoundSince"`
```
amoskyler: What does this property represent?

plugins/codeamp/gitsync_handler.go
```diff
@@ -167,6 +169,43 @@ func (x *CodeAmp) GitSyncEventHandler(e transistor.Event) error {
...
+				"project_id": project.ID,
+			})
+			return nil
+		}
+
+		// handle forced push commits
+		for _, feature := range features {
+			hash := feature.Hash
+			found := false
+			for _, commit := range payload.Commits {
```
amoskyler: Obviously performance won't be _too_ bad here, but what do you think about creating a `map[string]Commit` hash outside the features loop, so we can get O(1) lookups instead of looping through payload.Commits on every feature?

plugins/codeamp/gitsync_handler.go
```diff
@@ -167,6 +169,43 @@ func (x *CodeAmp) GitSyncEventHandler(e transistor.Event) error {
...
+				if commit.Hash == hash {
+					found = true
+				}
+			}
+			if !found {
```
amoskyler: Does payload.Commits hold every single commit in the tree for this commit? Based on the logic in gitsync it seems that it may only hold 50 of the commits.

gitsync.go#L174 
```
output, err = x.git(env, "-C", repoPath, "log", "--first-parent", "--date=iso-strict", "-n", "50", "--pretty=format:%H#@#%P#@#%s#@#%cN#@#%cd", git.Branch)
```

plugins/gitsync/gitsync.go
davechan813: Yes, should be deleted

plugins/events.go
davechan813: `NotSinceFor` means the time when this particular feature cannot be found in the latest github commits sync

plugins/codeamp/gitsync_handler.go
davechan813: sure, done

plugins/codeamp/db/paginator.go
```diff
@@ -163,7 +163,7 @@ func (r *FeatureListResolver) Entries() ([]*FeatureResolver, error) {
 	var results []*FeatureResolver
 
 	limit, offset := GetPaginatorLimitAndOffset(r.PaginatorInput)
-	err := r.DB.Order("created desc").Limit(limit).Offset(offset).Find(&rows).Error
+	err := r.DB.Where("not_found_since IS NULL").Order("created desc").Limit(limit).Offset(offset).Find(&rows).Error
```
davechan813: When querying features and sending to frontend, only send features that is not missing

plugins/codeamp/db/release.go
```diff
@@ -167,3 +167,13 @@ func (r *ReleaseResolver) Environment() (*EnvironmentResolver, error) {
...
 }
+
+// Redeployable
+func (r *ReleaseResolver) Redeployable() bool {
+	return r.Release.Redeployable
+}
+
+// RedeployableMessage
+func (r *ReleaseResolver) RedeployableMessage() string {
+	return r.Release.RedeployableMessage
```
davechan813: If a release is related to a missing feature, the `Redeloyable` field will be set to false

plugins/codeamp/gitsync_handler.go
```diff
@@ -167,6 +169,45 @@ func (x *CodeAmp) GitSyncEventHandler(e transistor.Event) error {
...
 		}
 
+		// put payload commits' hash in a map for quicker query below
+		payloadCommits := make(map[string]bool)
+		for _, commit := range payload.Commits {
+			payloadCommits[commit.Hash] = true
+		}
+
+		// limit the features array to 50, becaues payload.Commits is also of max length 50
+		if x.DB.Where("project_id = ? AND ref = ? AND not_found_since IS NULL", project.ID, fmt.Sprintf("refs/heads/%s", payload.Git.Branch)).Order("created desc").Limit(50).Find(&features).RecordNotFound() {
```
davechan813: When querying features list from DB, only query those on the same `ref` and can be found

plugins/codeamp/gitsync_handler.go
```diff
@@ -167,6 +169,45 @@ func (x *CodeAmp) GitSyncEventHandler(e transistor.Event) error {
...
+				"project_id": project.ID,
+			})
+			return nil
+		}
+
+		// handle forced push commits
+		for _, feature := range features {
+
+			// set NotFoundSince for the unfound feature
+			if _, ok := payloadCommits[feature.Hash]; !ok {
```
davechan813: If cannot found from the hash map

plugins/codeamp/graphql/feature.go
```diff
@@ -52,6 +53,14 @@ func (r *FeatureResolver) Created() graphql.Time {
...
 }
 
+// NotFoundSince
+func (r *FeatureResolver) NotFoundSince() graphql.Time {
+	if r.DBFeatureResolver.Feature.NotFoundSince == nil {
+		return graphql.Time{Time: time.Time{}}
+	}
+	return graphql.Time{Time: *r.DBFeatureResolver.Feature.NotFoundSince}
+}
+
```
davechan813: Change the type of time to `pointer` so that it can have null value. If pointer is `nil`, return a zero time value.

plugins/gitsync/gitsync.go
```diff
@@ -219,10 +219,6 @@ func (x *GitSync) Process(e transistor.Event) error {
 			c.Ref = fmt.Sprintf("refs/heads/%s", payload.Git.Branch)
 
 			_commits = append(_commits, c)
-
-			if c.Hash == payload.From {
-				break
-			}
```
davechan813: remove these lines so that the payload pulled from github will be complete, instead of just the part after last time pull, which is `payload.From`

plugins/codeamp/gitsync_handler.go
davechan813: good idea, changed

plugins/codeamp/db/paginator.go
davechan813: Yes. I release the feature `Add postgres ...`, then I removed it by `rebase`. The `release` page still looks fine.

Screen Shot 2019-08-05 at 3 40 07 PM
Screen Shot 2019-08-05 at 3 40 11 PM

PR #496, opened 2019-08-14
Check for pod restarts on deploy

aballman:
* Allows a budget of 1/4 of the pods requested worth of restarts before marking the build as failed.


plugins/kubernetes/deployment.go
```diff
@@ -1013,12 +1031,23 @@ func (x *Kubernetes) waitForDeploymentSuccess(clientset kubernetes.Interface,
 						continue
 					}
 
+					totalPodsForDeployment := len(allPods.Items)
+					totalPodRestarts := 0
+
+					allowedPodRestartsRatio := 0.25
```
drshrey: Can we have a comment for this? Why 0.25?

PR #498, opened 2019-09-10
Add machinebox/graphql dependency in vendor/

drshrey:
This dependency is being used within the Drone plugin to make requests to the CodeAmp API in order to get information about the environments a feature has been deployed into.


aballman:
Thanks for splitting this into a separate PR. Can you add a short description as to what this is used for? For Drone?

PR #506, opened 2019-10-30
Admins can see protected variables

aballman:
* Admins can see protected variables


plugins/codeamp/db/secret.go
```diff
@@ -19,7 +20,9 @@ func (r *SecretResolver) IsSecret() bool {
 // Value
 func (r *SecretResolver) Value() string {
 	if r.IsSecret() {
```
drshrey: Could we simplify to `if r.isSecret() && !r.isAdmin` vs. doing it nested?

PR #508, opened 2019-12-19
Bug fix for gokong/routes file. Add to upstream maybe?

aballman:
Fixes GoKong library when Listing Routes.
Seems to be a problem when interacting with Kong v1.0 with this version of the library.

Only seemed to break once we hit the pagination limit then got stuck in an infinite loop! :(


aballman:
Turns out just using the wrong version of this lib with kongv1 :(

PR #509, opened 2020-01-16
Add env vars for DataDog AMP

codeshrew:


plugins/kubernetes/deployment.go
```diff
@@ -852,6 +879,33 @@ func (x *Kubernetes) deployServices(clientset kubernetes.Interface,
...
+				Name: "DATADOG_TRACE_AGENT_HOSTNAME",
+				ValueFrom: &v1.EnvVarSource{
+					FieldRef: &v1.ObjectFieldSelector{
+						APIVersion: "v1",
+						FieldPath:  "spec.nodeName",
+					},
+				},
+			},
+			v1.EnvVar{
+				Name: "DD_AGENT_HOST",
```
drshrey: What's the difference between this and the HoSTNAME variable?

codeshrew: `DATADOG_TRACE_AGENT_HOSTNAME` and `STATSD_URL` are hardcoded into the applications, we're overriding them to make sure the old value doesn't remain.

`DD_AGENT_HOST` is the correct one https://docs.datadoghq.com/agent/docker/apm and is our new default going forward. It is also hardcoded in some of the applications, and so we are forcing it to the correct settings.

PR #520, opened 2020-04-01
Prevent deleting all services 

aballman:
Prevent deleteServices mutation from deleting all services if no ID or invalid ID is provided


itsmeniko:
looks good to me, but I do not have complete awareness of all the moving parts in the codeamp codebase. The changes included seem to address the bug we saw today, though! :tada:

